PR #9, opened 2019-09-26
Update requirements.md

rodrigopedra:
composer.json currently requires components from the 6.0 Illuminate components


freekmurze:
Thank you! 👍

PR #19, opened 2019-10-14
Fixed method name for replay detection

chinleung:
The correct method name for detecting replay should be `Spatie\EventProjector\Facades\Projectionist::isReplaying();`.


riasvdv:
Thanks!

PR #23, opened 2019-10-17
StoredEventRepository should not be opinionated

colinc:
The StoredEventRepository interface should not impose an int $startingFrom type as another repository implementation may use a different lookup method.

An example would be for a repository which does not use an auto-incrementing primary key and, instead of returning the events ordered by ID, returns the events ordered by milliseconds, microseconds or nanoseconds. In this implementation retrieveAllStartingFrom may expect a float, string, or some other variable type to appropriately handle the offset and replay the events.

### DIFF
--- a/src/EloquentStoredEventRepository.php
+++ b/src/EloquentStoredEventRepository.php
@@ -35,7 +35,7 @@ class EloquentStoredEventRepository implements StoredEventRepository
         });
     }
 
-    public function retrieveAllStartingFrom(int $startingFrom, string $uuid = null): LazyCollection
+    public function retrieveAllStartingFrom($startingFrom, string $uuid = null): LazyCollection
     {
         /** @var \Illuminate\Database\Query\Builder $query */
         $query = $this->storedEventModel::query()->startingFrom($startingFrom);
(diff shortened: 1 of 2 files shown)


freekmurze:
I'm thinking nearly everyone will use an incremeting id. Do you have an actual real world use case where you can't use an int?

colinc:
@freekmurze 

Yes. I'm actually going to be using this library in a situation where we'll have UUID ids instead of incrementing int ids as the data will be coming from and federated with more that one source so we can't synchronize or guarantee incrementing keys. Instead every server is using a highly accurate clock and the same leap smearing algorithm to ensure events can be reconstituted correctly by timestamp instead of by ID order.

So, because of that, when replaying events it requires that we use an offset "startingFrom" value that's larger than an int (as we'll be storing microseconds). Obviously this is a very fringe edge case scenario and, if necessary, we can fork the repo and maintain this ourselves. It would however be preferable to not go that route.

Either way, thank you for your time and the impressive number of libraries, packages, and contributions you and Spatie have made.

erikgaal:
Hi there,

We're using `Str::orderedUuid()` as our `event_id` which would still allow us to use _some_ `startingFrom` value, while allowing for parallel inserts in the database on different aggregates.

freekmurze:
Changing the interface is a breaking change. I'm not willing to create a new major version now, but I'll revisit this when working on v2.

PR #25, opened 2019-10-23
Implemented missing HandleDomainEventJob interface

colinc:
Both the config and documentation reference a missing Spatie\EventSourcing\HandleDomainEventJob class. At first I assumed this was a typo for the HandleStoredEventJob class, however, the class is final and can't be extended.

Instead, as hopefully makes more sense, I implement a HandleDomainEventJob interface in a non-breaking way that follows the rest of the codebase. I additionally updated the documentation to reference implementing rather than extending the new HandleDomainEventJob interface.


riasvdv:
Thanks!

PR #27, opened 2019-10-25
Use a UUID field when possible for storing UUIDs

colinc:
Using a UUID field, rather than a varchar, to store the UUIDs in Postgres has a storage and performance benefit (the UUID field only occupies [16 bytes](https://github.com/postgres/postgres/blob/master/src/include/utils/uuid.h#L18) as opposed to the 37 bytes the varchar is currently occupying).

Since the [MySQL grammar](https://github.com/illuminate/database/blob/master/Schema/Grammars/MySqlGrammar.php#L735) handles the lack of a native UUID type there is no negative impact on MySQL for this change.

This simply changes the stub and has no impact on existing codebases. I additionally updated the FakeUUID class to generate a valid V4 UUID so all the tests still pass on both MySQL and Postgres.

### DIFF
--- a/stubs/create_stored_events_table.php.stub
+++ b/stubs/create_stored_events_table.php.stub
@@ -10,7 +10,7 @@ final class CreateStoredEventsTable extends Migration
     {
         Schema::create('stored_events', function (Blueprint $table) {
             $table->increments('id');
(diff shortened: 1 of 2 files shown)


riasvdv:
Thanks!

PR #28, opened 2019-10-27
Use "self" typehints in AggregateRoot

spaceemotion:
While IDEs like PHPStorm can infer custom method calls due to usage of
"$this", actual code completion seems to only be working when the
correct method name is being called, as no auto-completion hints
appear when referencing the base "AggregateRoot" type.

I wasn't sure what to include in the changelog since subclasses could  change the return type and may have to be updated. Would this be a .1 or a .0.1 release then?

### DIFF
--- a/src/AggregateRoot.php
+++ b/src/AggregateRoot.php
@@ -13,7 +13,7 @@ abstract class AggregateRoot
     /** @var array */
     private $recordedEvents = [];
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this. I'm going to leave it like it at this point.

spaceemotion:
Are you planning an alternative solution for this? Without this fix, the project won't go through phpstan/larastan due to invalid typing.

PR #31, opened 2019-10-31
Provide docblocks to AggregateRoot class

spaceemotion:
This fixes the "method not found" errors in IDEs and static analysis tools like phpstan.

---

Followup for #28


freekmurze:
Thanks!

PR #33, opened 2019-11-20
Fix replay from specified event id

m-bymike:
Today I tried to replay a projection not from beginning of the event stream, but from a certain event id. This worked in the "old" event projector packages, but to my surprise it showed the following:

console

A quick look in the code revealed, that the wrong method was called... Here's the fix.

* Call correct method
* Organize params, as the first one now should be the starting id
* Remove coalesce operator (param has to be int and never is null)


freekmurze:
Thanks!

PR #37, opened 2019-12-06
Update installation-setup.md

kimuliSiraje:
```bash
composer require spatie/laravel-event-sourcing:1
```
👆was appearing on docs for V2


freekmurze:
Thanks!

PR #45, opened 2020-01-10
No dispatch of stored-event-job if no async projectors and no reactors for stored event

gdebrauwer:
If a stored event has no async / queued projectors and no reactors, a job is stil dispatched. This means a lot of useless jobs are dispatched.

This PR prevents the job from being dispatched if no async / queued projectors and no reactors are found for the stored event.


freekmurze:
Thanks!

PR #48, opened 2020-01-16
Snapshotting AggregateRoot

riasvdv:
This implements snapshotting of AggregateRoots
- Adds a `snapshot()` method to the AggregateRoot which creates a new Snapshot
- AggregateRoots track a `version` number, which increments each time an event is applied
- Persisting the events persists them with the version number
- The AggregateRoot has a `getState` and `useState` function to determine the state that is saved and retrieved from the snapshot, by default all properties on the class are stored in the snapshot

- When initializing an AggregateRoot, we check if there is a snapshot, if there is we load the state first, then retrieve and reconstitute from all events that happened after the snapshot.
- All public properties on an aggregate are stored as the state of the snapshot, in the tests this is the `$balance` property of the `AccountAggregate`, when restoring a snapshot, we set them back.

- We should also credit https://github.com/EventSaucePHP/EventSauce as the implementation is based on how EventSauce does it.

This will require a new major version, as some interfaces and a migration have changed, steps to upgrade:
- Add an `aggregate_version` property to the `stored_events` table `$table->unsignedInteger('aggregate_version')->nullable();`
- Republish the migrations or copy the `create_snapshots_table` migration
- The `StoredEventRepository` interface has a new method `retrieveAllAfterVersion` that you must implement if you have a custom repository
- The `StoredEventRepository` now accepts an `aggregateVersion` parameter in the `persist` and `persistMany` methods

### DIFF
--- a/src/AggregateRoot.php
+++ b/src/AggregateRoot.php
@@ -65,7 +65,7 @@ abstract class AggregateRoot
         return $this->getSnapshotRepository()->persist(new Snapshot(
             $this->aggregateUuid,
             $this->aggregateVersion,
-            $this->state(),
+            $this->getState(),
         ));
     }
 
@@ -84,14 +84,21 @@ abstract class AggregateRoot
         return $this->recordedEvents;
     }
 
-    private function state(): array
+    protected function getState(): array
     {
         $class = new ReflectionClass($this);
 
-        return collect($class->getProperties(ReflectionProperty::IS_PUBLIC))
+        return collect($class->getProperties())
             ->mapWithKeys(function (ReflectionProperty $property) {
                 return [$property->getName() => $this->{$property->getName()}];
             })->toArray();
+    }
+
+    protected function useState(array $state): void
+    {
+        foreach ($state as $key => $value) {
+            $this->$key = $value;
+        }
     }
 
     private function getAndClearRecordedEvents(): array
@@ -117,9 +124,8 @@ abstract class AggregateRoot
         }
 
         $this->aggregateVersion = $snapshot->aggregateVersion;
-        foreach ($snapshot->state as $key => $value) {
-            $this->$key = $value;
-        }
+
+        $this->useState($snapshot->state);
 
         $storedEventRepository->retrieveAllAfterVersion($this->aggregateVersion, $this->aggregateUuid)
             ->each(function (StoredEvent $storedEvent) {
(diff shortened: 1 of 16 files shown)


src/AggregateRoot.php
```diff
@@ -67,6 +84,16 @@ public function getRecordedEvents(): array
         return $this->recordedEvents;
     }
 
+    private function state(): array
+    {
+        $class = new ReflectionClass($this);
+
+        return collect($class->getProperties(ReflectionProperty::IS_PUBLIC))
```
freekmurze: Does this mean that if you want to use snapshots, all properties you want to use should be public?

src/AggregateRoot.php
```diff
@@ -78,7 +105,23 @@ private function getAndClearRecordedEvents(): array
 
     private function reconstituteFromEvents(): AggregateRoot
     {
-        $this->getStoredEventRepository()->retrieveAll($this->aggregateUuid)
+        $storedEventRepository = $this->getStoredEventRepository();
```
freekmurze: Maybe we should try to find and use a snapshot first. 

In either case we could call `retrieveAllAfterVersion`. If there was no snapshot that function will (or should) start getting all events, if there was a snapshot it should only get events after the snapshot (I guess that already works now)

stubs/create_snapshots_table.php.stub
```diff
@@ -0,0 +1,20 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+final class CreateSnapshotsTable extends Migration
```
freekmurze: Let's just drop all usage of `final`.

riasvdv: I'll do this in a separate PR

src/AggregateRoot.php
```diff
@@ -67,6 +84,23 @@ public function getRecordedEvents(): array
         return $this->recordedEvents;
     }
 
+    protected function getState(): array
```
freekmurze: Nice!

freekmurze:
Very nice work @riasvdv! I'll already merge this to v3. Could you also add docs for this on that branch?

warren32:
This is great thank you @riasvdv

PR #54, opened 2020-01-25
Do not allow concurrent persists by default

freekmurze:


src/AggregateRoot.php
```diff
@@ -6,19 +6,25 @@
...
 use ReflectionClass;
 use ReflectionProperty;
+use Spatie\EventSourcing\Exceptions\CouldNotPersistAggregate;
 use Spatie\EventSourcing\Snapshots\Snapshot;
 use Spatie\EventSourcing\Snapshots\SnapshotRepository;
 
 abstract class AggregateRoot
 {
-    private string $aggregateUuid;
+    private string $aggregateUuid = '';
```
riasvdv: Rename aggregateProperties without "aggregate", since we are in an aggregate class

src/AggregateRoot.php
```diff
@@ -31,7 +37,8 @@ public static function retrieve(string $uuid): AggregateRoot
     }
 
     /**
-     * @param  ShouldBeStored  $domainEvent
+     * @param ShouldBeStored $domainEvent
+     *
```
riasvdv: Do we need this docblock?

src/StoredEventRepository.php
```diff
@@ -17,4 +17,6 @@ public function persist(ShouldBeStored $event, string $uuid = null, int $aggrega
     public function persistMany(array $events, string $uuid = null, int $aggregateVersion = null): LazyCollection;
 
     public function update(StoredEvent $storedEvent): StoredEvent;
+
+    public function getLatestVersion(string $aggregateUuid): int;
```
riasvdv: Maybe we should rename this to `->getLatestAggregateVersion` so it's clear to someone implementing the interface that they need to return the aggregate's version

src/AggregateRoot.php
freekmurze: Rename `aggregateUuid` to `uuid`.

src/AggregateRoot.php
freekmurze: Removed!

src/StoredEventRepository.php
freekmurze: Good idea. Done!

PR #56, opened 2020-01-29
Fix testing aggregates docs

jamesfairhurst:
Thanks for the great packages, fixed a few minor issues I ran into when going through the testing docs. Hopefully makes them a little better.


freekmurze:
Thanks!

PR #57, opened 2020-01-29
add down() to CreateStoredEventsTable

4n70w4:


freekmurze:
Thanks, but I generally don't add `down` functions to my migrations.

4n70w4:
It is necessary for successful execution
`php artisan migration:rollback && php artisan migration`
without errors.

riasvdv:
You can add this function in your own migrations, we generally only use `migrate:fresh`

4n70w4:
The package description does not say what to use `migrate:fresh` instead `migration:rollback && migration`.

4n70w4:
I see no objective reason to reject this PR.

PR #58, opened 2020-02-05
fix slow replay

m-bymike:
LazyCollections count implementation iterates over all results. The replay command counts twice, so we end up iterating three times. By moving the count operation to the database and using the result in a variable we are able to get down to a single iteration.

Our system currently has around 300'000 stored events. To replay a projection, the command looped 600'000 times only to check if there are events and to display the progress bar, delaying the actual replay by several minutes.

This PR is aiming at version 1, because the upgrade to PHP 7.4 is planned later this year.


freekmurze:
Thank you, I'll take a look at this soon, and will port it to v2 and v3.

freekmurze:
Thanks!

PR #59, opened 2020-02-07
Publish new snapshots table

SiebeVE:


freekmurze:
Thanks! Happy snapshotting!

PR #62, opened 2020-02-14
Only replace the first instance of the basePath

hailwood:
We use docker containers,
Our project sits at `/app` in the filesystem, meaning we have paths like `/app/app/Http/Controller.php`

With the current setup it was meaning the namespace of `\App\Http\Controllers\Controller` was becoming `\Http\Controllers\Controller`


freekmurze:
Thanks!

PR #67, opened 2020-02-22
add tests for datetime deserialization

dilab:
Run test as:
`vendor/bin/phpunit --filter EventSerializerTest`

You will notice the bug:

```
1) Spatie\EventSourcing\Tests\EventSerializers\EventSerializerTest::it_can_deserialize_an_event_with_datetime
Symfony\Component\Serializer\Exception\MissingConstructorArgumentsException: Cannot create an instance of DateTimeImmutable from serialized data because its constructor requires parameter "time" to be present.
```


freekmurze:
Very nice thanks!

PR #75, opened 2020-03-21
expose AggregateRoot for testing state

dilab:
added a getter method to FakeAggregateRoot


freekmurze:
I've cleaned it up.

Thanks!

PR #80, opened 2020-04-10
start Mysql service before creating db for github action

sohaibafifi:
Signed-off-by: Sohaib LAFIFI <sohaib.lafifi@univ-artois.fr>


freekmurze:
Thanks!

PR #81, opened 2020-04-12
Fix missing word in installation setup docs

jessarcher:
Came across this while installing the package. Thank you for everything you do :heart:


freekmurze:
Thanks!

PR #83, opened 2020-04-14
Fixed event replay command docs

joshralph93:
The docs currently reference that the `--projector` option is available when replaying events. This has actually been changed to an option, therefore no `--projector` prefix required.

Looks like this change was introduced [here](https://github.com/spatie/laravel-event-sourcing/commit/fbb439093285522145ec4c9a8e5e70c268b4a0c7#diff-219d2a262b5ca12ebf45af61b8915600)


freekmurze:
Thanks!

PR #85, opened 2020-04-15
Remove the Final keyword from code generation examples

pascualstromsnes:
I removed the Final keyword from the examples in the docs for using the CLI to generate aggregates as the stubs no longer uses Final.


freekmurze:
Thanks!

PR #86, opened 2020-04-16
Add a --force flag to the replay command

pascualstromsnes:
This PR adds a force flag / option to the replay command, as well as a corresponding test.

### DIFF
--- a/src/Console/ReplayCommand.php
+++ b/src/Console/ReplayCommand.php
@@ -12,7 +12,8 @@ class ReplayCommand extends Command
 {
     protected $signature = 'event-sourcing:replay {projector?*}
                             {--from=0 : Replay events starting from this event number}
-                            {--stored-event-model= : Replay events from this store}';
+                            {--stored-event-model= : Replay events from this store}
+                            {--force : Replay events without user prompt}';
 
     protected $description = 'Replay stored events';
 
@@ -35,7 +36,7 @@ class ReplayCommand extends Command
 
     public function selectProjectors(array $projectorClassNames): ?Collection
     {
-        if (count($projectorClassNames ?? []) === 0) {
+        if (count($projectorClassNames ?? []) === 0 && ! $this->option('force')) {
             if (! $confirmed = $this->confirm('Are you sure you want to replay events to all projectors?', true)) {
                 return null;
             }
--- a/tests/Console/ReplayCommandTest.php
+++ b/tests/Console/ReplayCommandTest.php
@@ -68,7 +68,7 @@ class ReplayCommandTest extends TestCase
             ->assertExitCode(0);
     }
 
-    //**/*@test*/*//
+    /** @test */
     public function it_will_replay_all_events_without_user_prompt_if_the_force_option_is_given()
     {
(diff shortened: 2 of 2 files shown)


tests/Console/ReplayCommandTest.php
```diff
@@ -68,6 +68,16 @@ public function if_no_projectors_are_given_it_will_ask_if_it_should_run_events_a
             ->assertExitCode(0);
     }
 
+    //**/*@test*/*//
```
50bhan, line 71: Why the annotation is like this?

pascualstromsnes: Seems like my IDE did something funny to the formatting of the annotation, fixed now. Thank you!

src/Console/ReplayCommand.php
```diff
@@ -35,7 +36,7 @@ public function handle(Projectionist $projectionist): void
 
     public function selectProjectors(array $projectorClassNames): ?Collection
     {
-        if (count($projectorClassNames ?? []) === 0) {
+        if (count($projectorClassNames ?? []) === 0 && ! $this->option('force')) {
```
freekmurze, line 39: Instead of using `&&` just add the check above the existing code.

pascualstromsnes: Sure thing, that’ll be way more readable. Will post a fix later today.

pascualstromsnes: Sorry about the delay @freekmurze 

Here is what I had in mind for a refactor of my first suggestion:

```php
public function selectProjectors(array $projectorClassNames): ?Collection
    {
        if (count($projectorClassNames ?? []) === 0) {

            if ($this->option('force')) {
                return $this->projectionist->getProjectors();
            }

            if (! $confirmed = $this->confirm('Are you sure you want to replay events to all projectors?', true)) {
                return null;
            }

            return $this->projectionist->getProjectors();

        }

  // ...
```

To prevent a bunch of commits here, I wanted to hear if this seems like a good approach.

I have to admit, this is mye very first ever PR to an open source project (not counting documentation patches), so I am trying to thread carefully here. 
I also see that you prefer to have multiple commits in a PR squashed before submitting. I have never tried that before, but I could give it a try here if it is possible for a PR that has already been submitted.

freekmurze: Don't worry about too many commits, I'll squash them all myself before merging 👍 

Thanks for contributing!

Instead of checking the `force` flag yourself, use Laravels `ConfirmatbleTrait` which checks this for your. Here an example of how you can use it: https://github.com/spatie/laravel-db-snapshots/blob/5c1b58da6d1898eb5d2a4b31d6287d27b86f8188/src/Commands/Load.php#L13

freekmurze:
Closing due to inactivity.

pascualstromsnes:
Sorry about my tardiness here, a couple of exams got complete hold of my time and energy.
I will look into patching my PR as per your suggestions next week, @freekmurze

freekmurze:
No worries, there are no deadlines in open source.

I hope your exams went well!

PR #90, opened 2020-04-20
fixed typo

JayAhrDe:
A PR seems to be sort of an overkill for a tiny duplication, but as I just stumbled over it.


freekmurze:
Thanks!

PR #93, opened 2020-04-30
Add applied events support

brendt:
Calling `persist` on an aggregate root will clear the `recordedEvents` array. This means that if you're calling persists within a faked aggregate root, assertions will fail because the recorded events array is empty. 

I'd suggest to keep track of all applied events as well, which will never get cleared.


brendt:
These tests were already failing before this feature btw.

freekmurze:
Perfect, thanks!

PR #99, opened 2020-05-11
Update UPGRADING.md

immeyti:
In the middle of the upgrading process, I've encountered `Spatie\EventSourcing\AggregateRoot::getSnapshotRepository() must be an instance of Spatie\EventSourcing\Snapshots\SnapshotRepository, instance of Laravel\Lumen\Application returned` error, so I found out some configs are missing.


freekmurze:
Thanks!

PR #101, opened 2020-05-15
Relax property type of meta_data

### LINKED ISSUE #100: SchemalessAttributes needed?
Hi! Love the package! 

I'm building custom repositories to store and load events from https://eventstore.com/ so I'm not using Eloquent at all. 

I liked the use of StoredEvent, except for the SchemalessAttributes requirement. This type then enforces all of Eloquent to be brought in. 

https://github.com/spatie/laravel-event-sourcing/blob/ea48283a7a0db3a5c090de799df8cf01f5918f69/src/StoredEvent.php#L24

At the moment I'm using a workaround, but it's not pretty!

```php
        use Illuminate\Database\Eloquent\Model;
        $emptyModel = new class extends Model { };
        $model = new $emptyModel();
        $model->meta_data = $metadata;

        return new StoredEvent([
            ...
            'meta_data' => new SchemalessAttributes($model, 'meta_data'),
            ...
        ]);
```

morrislaptop:
Fixes #100

### DIFF
--- a/src/StoredEvent.php
+++ b/src/StoredEvent.php
@@ -21,7 +21,8 @@ class StoredEvent implements Arrayable
 
     public string $event_class;
 
-    public SchemalessAttributes $meta_data;
+    /** @var array|string */
+    public $meta_data;
 
     public string $created_at;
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you!

PR #102, opened 2020-05-18
Added rejection clause to getState for snapshots

roelofjan-elsinga:
The rejection clause was included in the source code, but not in the documentation.


freekmurze:
Thanks!

PR #105, opened 2020-05-20
Only include public properties of the aggregate when snapshotting

morrislaptop:
Otherwise properties like `aggregateVersion` and `aggregateVersionAfterReconstitution` are getting into the event properties.


freekmurze:
Makes sense, thanks!

PR #109, opened 2020-06-09
Use Laravel container to call aggregate apply method

### LINKED ISSUE #108: Method injection for aggregate apply* methods like reactors and projectors
Would you accept a PR for this?

morrislaptop:
Fixes #108 

This change allows the `apply` methods on an aggregate to use method injection, so you can resolve any dependencies you need in those methods as well. This is in line with projectors and reactors.

It does require the apply method to call the variable `$event`, however this is a convention that is used for projections and reactors so I don't see this being a problem. 

If method injection fails for whatever reason, it resorts to the original approach so there shouldn't be any breaking changes. 

```php
<?php

namespace Spatie\EventSourcing\Tests\TestClasses\AggregateRoots;

use Math;
use Spatie\EventSourcing\AggregateRoot;
use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\StorableEvents\MoneyAdded;
use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\StorableEvents\MoneyMultiplied;

class AccountAggregateRoot extends AggregateRoot
{
    public int $balance = 0;

    public function multiplyMoney(int $amount): self
    {
        $this->recordThat(new MoneyMultiplied($amount));

        return $this;
    }

    public function applyMoneyMultiplied(MoneyMultiplied $event, Math $math)
    {
        $this->balance = $math->multiply($this->balance, $event->amount);
    }
}

```

### DIFF
--- a/.github/workflows/php-cs-fixer.yml
+++ b/.github/workflows/php-cs-fixer.yml
@@ -1,0 +1,29 @@
+name: Check & fix styling
+
+on: [push]
+
+jobs:
+    style:
+        runs-on: ubuntu-latest
+
+        steps:
+            -   name: Checkout code
+                uses: actions/checkout@v1
+
+            -   name: Fix style
+                uses: docker://oskarstark/php-cs-fixer-ga
+                with:
+                    args: --config=.php_cs --allow-risky=yes
+
+            -   name: Extract branch name
+                shell: bash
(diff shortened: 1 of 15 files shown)


freekmurze:
Thank you for this! I'm going treat this as a breaking change, because of the added requirement to call that event variable `$event`

I'll merge this on the `v5` branch.

PR #112, opened 2020-06-18
Pass metadata to serializer to allow events to be upgraded

morrislaptop:
This is a first pass suggestion on how to implement an upgrade backwards compatible event versions. 

For example if we had an event like `AccountCreated` and stored the open date as `2019-03-43 12:30:00`. This event payload is v1. 

In later versions of the app, we're going international and so the open date now needs a timezone attached to it like `2019-03-43 12:30:00 +00:00`.

We change the code to include the timezone and it's now v2 of the event payload. 

When we replay events we expect the latest version of the event payload in order to function correctly. This PR allows the event serializer to upgrade the event live so it can be transformed to the latest version.

### DIFF
--- a/docs/advanced-usage/using-your-own-event-serializer.md
+++ b/docs/advanced-usage/using-your-own-event-serializer.md
@@ -18,6 +18,31 @@ interface EventSerializer
 {
     public function serialize(ShouldBeStored $event): string;
 
-    public function deserialize(string $eventClass, string $json): ShouldBeStored;
+    public function deserialize(string $eventClass, string $json, string $metadata): ShouldBeStored;
 }
 ```
+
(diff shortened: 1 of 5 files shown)


freekmurze:
Changing the signature of an interface is a breaking change, so we can't pull this in now. I'll revisit this PR again when creating a new major version.

morrislaptop:
I did target v4, it would be good to have a section in the docs with an approach on upgrading events

freekmurze:
@morrislaptop right, we could already pull this in v4. Could you add docs for the added functionality?

morrislaptop:
@freekmurze done :)

freekmurze:
Thanks!

Issue #114, opened 2020-06-23
Expose aggregate_version of StoredEvent


dilab:
Can we expose aggregate_version of StoredEvent class. It will enable us to use the version when building a projection. This is useful when two events happened at the same time.

freekmurze:
I'd accept a PR for this.

dilab:
@freekmurze done https://github.com/spatie/laravel-event-sourcing/pull/115.

freekmurze:
Thanks, we'll continue the conversation in #115

PR #115, opened 2020-06-23
expose aggregate_version of StoredEvent

dilab:
As discussed in https://github.com/spatie/laravel-event-sourcing/issues/114


freekmurze:
Could you add a test that covers the added functionality?

dilab:
It is being tested by other tests.

freekmurze:
By which one?

The test that I am looking for should fail when the code of this PR should be removed.

dilab:
@freekmurze test added. please review.

freekmurze:
Thanks!

PR #118, opened 2020-07-03
Fixed the docs for addReactors and addProjectors

roelofjan-elsinga:
The singular method name was used, while this should be the plural method name.


freekmurze:
Nice catch, thanks!

PR #120, opened 2020-07-09
Documentation corrections

dlaytonjames:
Documentation corrections
1. Example config changes - base model name, EventSerializer Interface name in the comment
2. Package name in Questions and Issues


freekmurze:
Thank you!

PR #123, opened 2020-07-14
Replace model where clause with uuid model scope

dlaytonjames:
Make EloquentStoredEventRepository consistently use the uuid model scope for queries.

This allows for overriding the scope for use cases such as using the efficientUuid package


freekmurze:
Thanks!

PR #124, opened 2020-07-14
Default to BigIncrements on package table stubs

dlaytonjames:
Ensures many more events and snapshots can be recorded before an int overflow error occurs.

Also bring it inline with the default int type for Laravel 7 id column types.

Regular unsigned int is not large enough for event heavy Event Sourced systems.


freekmurze:
Nice, thanks!

PR #127, opened 2020-07-16
add SyncReactor

dilab:
By default **Reactor** handles events asynchronously. However, in some situations, we need to do things synchronously. For example, in an **account** domain, when money is deducted, we want to deduct money from **wallet** domain. This needs to be done synchronously to ensure its consistency. 

In this pull request, we create a new interface **SyncReactor**, when a reactor implements this interface, it will handle events synchronously.


freekmurze:
The tests seems to be failing for this one, could you take a look at that?

Could you also update the docs with a description + example of the new functionality?

Thanks!

dilab:
fixed test.

dilab:
added doc

freekmurze:
Thank you for your work on this.

We're not going to merge this in. This PR inspired us to rethink how projector and reactors work. In the upcoming v4 of this package both projector and reactors will be handle sync by default. If a projector or reactor should be handled async, the ` Illuminate\Contracts\Queue\ShouldQueue` marker interface should be implement on the class.

dilab:
@freekmurze hmmm, any timeline for release of v4?

freekmurze:
Somewhere in the next few weeks

PR #130, opened 2020-07-20
Support for Lumen

VeeeneX:
Lumen doesn't have function to get app_path(), but there are following workarounds:

```php
app()->path();
app('path');
base_path('app');
```


freekmurze:
Make this change in the config file of your own project.

PR #131, opened 2020-07-27
Updates implementation of how to queue a projector

karlos545:
Whilst pulling down v4 and having a play around I realised at the queuing section of the documentation was still at v2. If I find anything else I'll drop it in this morning. I won't be offended if you want to reword it 👍


freekmurze:
Thanks!

PR #132, opened 2020-07-27
Update snapshots.md to reflect current code

m40l:
This was changed in the code in https://github.com/spatie/laravel-event-sourcing/commit/2e927470914a65c19a33037d1f3e38508c78fff3 but the change was not reflected in this line in the docs.


freekmurze:
Thanks!

PR #134, opened 2020-08-03
Add assertion that specific event is recorded

Robertbaelde:
Sometimes you'd like to test if a specific event is fired from the method of an aggregate, without caring about other events that the method might fire. For this, I've added a `assertEventRecorded` to the fake aggregate root. 

So we could use: 

`->assertEventRecorded(new DummyEvent(3));`

While `DummyEvent(4)` is also fired from the same method.


freekmurze:
Thanks!

PR #135, opened 2020-08-06
Allow events to be dispatched from an aggregate root

Robertbaelde:
Currently when an event is recorded from within an aggregate root, it is not dispatched on Laravel's event bus. 

 This non breaking, configurable change would allow for aggregate roots to dispatch the events when the events are persisted.

### DIFF
--- a/config/event-sourcing.php
+++ b/config/event-sourcing.php
@@ -107,8 +107,9 @@ return [
     'cache_path' => storage_path('app/event-sourcing'),
 
     /*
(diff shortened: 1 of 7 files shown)


freekmurze:
Thanks for your work on this!

Could you take care in the failing tests in Laravel 6?

Robertbaelde:
Yes, on it! Trying to replicate as we speak.

freekmurze:
Thank you!

PR #136, opened 2020-08-06
Allow protected apply methods

sebastiandedeyne:
`app()->call()` currently fails when a method is `protected`. 

`method_exists` returns `true` for protected methods here (probably because the check happens in the parent `AggregateRoot` class) When the protected methods get called, there's an exception because `app()->call()` wants a public method.


freekmurze:
Nice catch!

Could you also add a test?

src/AggregateRoots/AggregateRoot.php
```diff
@@ -195,12 +195,19 @@ private function apply(ShouldBeStored $event): void
...
 
-        if (method_exists($this, $applyingMethodName)) {
+        $reflectionClass = new ReflectionClass($this);
+
+        $applyMethodExists = $reflectionClass->hasMethod($applyingMethodName);
+        $applyMethodIsPublic = $applyMethodExists && $reflectionClass->getMethod($applyingMethodName)->isPublic();
+
+        if ($applyMethodExists && $applyMethodIsPublic) {
             try {
                 app()->call([$this, $applyingMethodName], ['event' => $event]);
```
riasvdv, line 205: What's the benefit of using `app()->call()`? Can't we just always do `$this->$applyingMethodName($event);` instead?

sebastiandedeyne: `app()->call()` allows the package user to inject additional dependencies in the `apply*` methods.

I don't see much value in that feature, but removing it now would be a breaking change.

riasvdv: Ah right, forgot about that

sebastiandedeyne:
I added a test for the DI behaviour, which I believe we should reconsider in the next major version.

sebastiandedeyne:
Rebased the branch to fix the tests. Anything else that needs to happen before we can merge?

freekmurze:
Thanks!

PR #137, opened 2020-08-13
Support Carbon dates in events

erikgaal:
Using `Carbon` dates in events doesn't currently work out of the box. The current implementation of the DateTimeNormalizer only allows you to use PHP's `DateTime` which misses out on some useful Carbon functionalities. To use Carbon dates, you need to implement your own [`EventSerializer`](https://github.com/spatie/laravel-event-sourcing/blob/master/src/EventSerializers/EventSerializer.php) and add a `Normalizer` that supports the (de)normalization of Carbon instances.

As the standard date implementation is Carbon and this package is built for Laravel, I would like to suggest to include a `CarbonNormalizer` to improve the developer experience for those who use Carbon datetimes in their events.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "spatie/laravel-schemaless-attributes": "^1.6",
         "symfony/finder": "^4.2|^5.0",
         "symfony/property-access": "^4.0|^5.0",
-        "symfony/serializer": "^4.0|^5.0"
+        "symfony/serializer": "^5.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.16",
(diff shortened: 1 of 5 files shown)


erikgaal:
I see that the tests are failing due to the difference in types between `symfony/serializer` v4 and v5. If this functionality is to be accepted, what would be the preferred course of action to fix these?

riasvdv:
Does it fail because symfony/serializer v4 doesn't support what you've added? In that case, you can add a min requirement to the composer.json for v5 I think

erikgaal:
Hi @riasvdv, wouldn't that impose a breaking change?

riasvdv:
@freekmurze Is requiring min symfony/serializer v5 a breaking change if prefer-stable already uses it?

freekmurze:
I don't think it is a breaking change.

freekmurze:
Thank you!

PR #138, opened 2020-08-17
add down routine to snapshots and stored_events migration stubs

dkerner:
without those down routines `php artisan migrate:refresh` will fail since the tables for `stored_events` and `snapshots` are already present during migration


riasvdv:
You can use `php artisan migrate:fresh` which will start with a clear database, we don't add `down` methods to our migrations.

PR #141, opened 2020-08-27
Add stored event repository as parameter in replay command

joricksteenstra:
I wanted to use the --stored-event-model option but I noticed that it wasn't used anymore in the replay command. It also made more sense to specify a stored event repository as option in the command instead of a stored event model. Added this because sometimes you would like to replay only the events from a specific store.


freekmurze:
Thank you for this. It's a breaking change, I'll revisit it when creating a new major version.

PR #144, opened 2020-09-05
Fix link to v1 docs in README

nathanheffley:
I keep using this link instead of the one at the top of the README for some reason and figured I may as well fix it!


freekmurze:
Thanks!

PR #146, opened 2020-09-08
Update documentation to match the code

jessarcher:
Updates a presumably outdated reference to an empty interface that doesn't exist, so that it now matches the code samples immediately below :smiley:


freekmurze:
Noce catch, thanks!

PR #149, opened 2020-09-15
Configurable event retrieval ordering

inxilpro:
Currently, the `EloquentStoredEventRepository` orders all retrieval queries by ID. This can be a problem in cases where there's no guarantee that events will be inserted exactly in the order that they were created, or cases where you want to back-fill events from a legacy system.

This PR adds a new configuration option, `stored_event_order_by` which is set to `id` by default. For some users, setting this to `created_at` may be preferable (as in our case).

While this doesn't entirely resolve the issues mentioned in #23, it may help in those cases without changing the interface of the repository.


freekmurze:
Thanks for your work on this. I don't think too many people need this, so I'm going to pass for now in order to keep the package simpele.

For your project you could use a `StoredEventRepository` of your own that extends the Eloquent one in the package and make the necessary changes on that one.

PR #153, opened 2020-09-17
Model and Collection normalization

inxilpro:
The current implementation that uses `SerializesModels` does not work with collections of models, and also fails if you type hint your StoredEvent's properties with the model class or `Collection`. This solves that by introducing a new `ModelIdentifierNormalizer` instead.

Under the hood, it defers to the same logic that `SerializesModels` uses, but it does it as part of the serialization/normalization chain.

I've also made the normalization chain configurable, so that it's not necessary to re-implement the serializer just to add custom normalizers. I'm happy to remove that customization if you feel it's too much, but I do think it's worthwhile.

This could be considered a breaking change, but I don't think it is—if `SerializesModels` exists on the event, it will still work correctly.

### DIFF
--- a/config/event-sourcing.php
+++ b/config/event-sourcing.php
@@ -89,6 +89,19 @@ return [
      * should implement Spatie\EventSourcing\EventSerializers\EventSerializer.
      */
     'event_serializer' => Spatie\EventSourcing\EventSerializers\JsonEventSerializer::class,
+    
+    /*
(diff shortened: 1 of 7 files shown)


freekmurze:
This seems like a quality PR. Thanks for your work on it. I'll review and probably release it tomorrow. 👍

freekmurze:
Thanks!

PR #154, opened 2020-09-18
Support closures in assert recorded

### LINKED ISSUE #150: Support closures in assertRecorded
It's currently difficult to use data generated in aggregate roots in assertions.

```php
$someUuid = null;

MyAggregateRoot::fake()
    ->when(function (MyAggregateRoot $myAggregateRoot) use (&$someUuid): void {
        $someUuid = $myAggregateRoot->doSomething();
    })
    ->assertRecorded([
        new AnEvent($someUuid),
    ]);
```

This would be easier if we could pass down data from `when`, to be used in a future closure.

```php
MyAggregateRoot::fake()
    ->when(function (MyAggregateRoot $myAggregateRoot): void {
        return $myAggregateRoot->doSomething();
    })
    ->assertRecorded(fn (SomeUuid $someUuid) => [
        new AnEvent($someUuid),
    ]);
```

brendt:
closes #150

### DIFF
--- a/src/AggregateRoots/FakeAggregateRoot.php
+++ b/src/AggregateRoots/FakeAggregateRoot.php
@@ -43,7 +43,7 @@ class FakeAggregateRoot
         return $this;
     }
 
-    public function assertThat(callable $callback): self
+    public function then(callable $callback): self
     {
         $result = $callback($this->whenResult);
 
--- a/src/AggregateRoots/FakeAggregateRoot.php
+++ b/src/AggregateRoots/FakeAggregateRoot.php
@@ -77,7 +77,7 @@ class FakeAggregateRoot
 
     public function assertNotRecorded($unexpectedEventClasses): self
     {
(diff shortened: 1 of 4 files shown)


brendt:
Giving it some thought, I think it's better to keep `assertRecorded` and `assertNotRecorded` as is, and only add `assertThat`. What's your opinion?

src/AggregateRoots/FakeAggregateRoot.php
```diff
@@ -36,7 +38,18 @@ public function given($events): self
...
+        $this->whenResult = $callable($this->aggregateRoot);
+
+        return $this;
+    }
+
+    public function assertThat(callable $callback): self
+    {
+        $result = $callback($this->whenResult);
+
+        if ($result !== null) {
```
brendt, line 50: I explicitly only assert when a result is returned. If no result is returned and no other assertions are performed — with `assertThat` or another assert function — PHPUnit will warn you that this test case is not doing any assertions. I think this is better than having the `$callback` result default to `true`.

freekmurze (approved):
Nice!

PR #156, opened 2020-09-28
Added loadUuid function to avoid static retrieve function

joricksteenstra:
This non-breaking PR will add a non-static method to load in the UUID so that the AggregateRoot can be injected.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -43,6 +43,13 @@ abstract class AggregateRoot
         return $aggregateRoot->reconstituteFromEvents();
     }
 
+    public function loadUuid(string $uuid): self
+    {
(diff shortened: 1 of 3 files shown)


freekmurze:
Nice, could you also add an example of this in the docs?

freekmurze:
Thanks!

src/AggregateRoots/AggregateRoot.php
```diff
@@ -43,6 +43,13 @@ public static function retrieve(string $uuid)
         return $aggregateRoot->reconstituteFromEvents();
     }
 
+    public function loadUuid(string $uuid): self
+    {
+        $this->uuid = $uuid;
```
Robertbaelde, line 48: Maybe there should be a guard here to protect an already loaded Aggregate to be re-loaded? 🤔

PR #158, opened 2020-10-01
Re-use existing instance of ShouldBeStored when possible

inxilpro:
Currently, when you call `EloquentStoredEventRepository::persist()`, this persists the event to the database and then immediately returns the deserialized version of that event. When the attributes on the event are all basic data types like `string` and `int`, the performance hit associated with serializing and deserializing is negligible. When one or more attributes on the event are eloquent models, this can result in an N+1 issue.

This PR introduces the concept of the "original event" to `StoredEvent` and `EloquentStoredEvent`. If the original event is provided and/or present, these models can re-use it. If not, the models can instantiate it from the serialized data.


freekmurze:
Thanks!

PR #159, opened 2020-10-01
Fix Paths and Reference URL in event-sourcing.php config file

dlaytonjames:
fix 'should implement' paths/namespaces for current version
Change reference url for caching event handlers to current v4 version url in the docs


dlaytonjames:
FYI, not a hacktoberfest PR ;)

freekmurze:
Thanks!

PR #160, opened 2020-10-07
Allow for snake_case parameters in handle methods

inxilpro:
This may be controversial, but I'll give it a shot :)

Our coding standard requires that variables be `$snake_cased` (like the argument for using a light editor theme, [there's evidence that it takes about 20% longer to read `$camelCase` variables](https://whatheco.de/2013/02/16/camelcase-vs-underscores-revisited/)).

This PR just adds `stored_event` and `aggregate_uuid` to the list of parameters available to event handler method injection. This gives the consuming code the option to use either style as they see fit!

### DIFF
--- a/src/EventHandlers/HandlesEvents.php
+++ b/src/EventHandlers/HandlesEvents.php
@@ -27,7 +27,9 @@ trait HandlesEvents
         $parameters = [
             'event' => $storedEvent->event,
             'storedEvent' => $storedEvent,
+            'stored_event' => $storedEvent,
             'aggregateUuid' => $storedEvent->aggregate_uuid,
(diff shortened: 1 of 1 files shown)


freekmurze:
To me,  use both cases there seems kinda dirty.

I'd accept a PR that moves [L27-L31](https://github.com/spatie/laravel-event-sourcing/pull/160/files#diff-3917cab66732578db4ff119e098c0bc9L27-L31) to a method called `getEventHandlerParameters` on the `StoredEvent` model. 

This way you could use a custom `StoredEvent` model that use a different implementation of `getEventHandlerParameters`.

freekmurze:
Closing due to inactivity.

PR #162, opened 2020-10-20
Add beforeApply to AggregateRoot

sebastiandedeyne:
This PR adds a `beforeApply` method to the `AggregateRoot` that can be overridden to _always_ do something before applying an event.

We often come across a situation where the aggregate root needs to be kickstarted with a `start` or `initialize` event of sort, and every other method needs to ensure that the aggregate root has been initialized.

```php
class PaymentAggregateRoot
{
    public int $due;

    public int $paid = 0;

    public function start(int $due): void
    {
    	$this->recordThat(new PaymentStartedEvent($due));
    }

    public function pay(int $amount): void
    {
    	if (! isset($this->due)) {
    		throw new Exception("A payment must always have a due amount");
    	}

    	$this->recordThat(new PaidEvent($amount));
    }
}
```

With `beforeApply`, we can move that rule to always be called, and not worry about it anymore.

```php
class PaymentAggregateRoot
{
    public int $due;

    public int $paid = 0;

    public function start(int $due): void
    {
    	$this->recordThat(new PaymentStartedEvent($due));
    }

    public function pay(int $amount): void
    {
    	$this->recordThat(new PaidEvent($amount));
    }

    public function beforeApply(ShouldBeStored $event): void
    {
        if ($event instanceof PaymentStartedEvent) {
            return;
        }

    	if (! isset($this->due)) {
    		throw new Exception("A payment must always have a due amount");
    	}
    }
}
```

Alternatively, we could make `apply` protected instead of private, so you could override it and call `parent::apply()`. I actually prefer this approach, but not sure if it has other implications. And can we do that non-breaking?

```php
class PaymentAggregateRoot
{
    public int $due;

    public int $paid = 0;

    public function start(int $due): void
    {
    	$this->recordThat(new PaymentStartedEvent($due));
    }

    public function pay(int $amount): void
    {
    	$this->recordThat(new PaidEvent($amount));
    }

    protected function apply(ShouldBeStored $event): void
    {
    	if ($event instanceof PaymentStarted) {
    		parent::apply($event);

    		return;
    	}

    	if (! isset($this->due)) {
    		throw new Exception("A payment must always have a due amount");
    	}

    	parent::apply($event);
    }
}
```

Ping @brendt for a second opinion.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -196,6 +196,8 @@ abstract class AggregateRoot
 
     private function apply(ShouldBeStored $event): void
     {
+        $this->beforeApply($event);
+
         $classBaseName = class_basename($event);
 
         $camelCasedBaseName = ucfirst(Str::camel($classBaseName));
@@ -213,6 +215,10 @@ abstract class AggregateRoot
         $this->appliedEvents[] = $event;
 
         $this->aggregateVersion++;
+    }
+
+    protected function beforeApply(ShouldBeStored $event): void
+    {
     }
 
     public static function fake(string $uuid = null): FakeAggregateRoot
--- a/tests/AggregateRootTest.php
+++ b/tests/AggregateRootTest.php
@@ -12,6 +12,7 @@ use Spatie\EventSourcing\Snapshots\EloquentSnapshot;
 use Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEvent;
 use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\AccountAggregateRoot;
 use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\AccountAggregateRootThatAllowsConcurrency;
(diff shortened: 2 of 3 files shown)


freekmurze:
> Alternatively, we could make apply protected instead of private, so you could override it and call parent::apply(). I actually prefer this approach, but not sure if it has other implications. And can we do that non-breaking?

I think this change would be non-breaking.

brendt:
I'm not sure what approach has my preference. Both feel like a workaround for not being able to override the constructor to ensure the aggregate root is valid. I realise there's not much we can do about that though unless we drastically change the design of aggregate roots, which I don't think is a good solution either.

I think `beforeApply` is more explicit, but that might actually be a bad thing in this case, since it's already an obscure feature. So honestly I'm a little torn about this one. I guess making the method protected is the easiest solution.

morrislaptop:
FWIW when implementing https://github.com/spatie/laravel-event-sourcing/pull/182 in my project, I hit a big roadblock with `apply` being protected.

brendt:
I think this feature isn't relevant anymore since we're using internal state machines/internal states.

PR #171, opened 2020-11-21
Add persistWithTransaction method to aggregate root

morrislaptop:
Sometimes I might want to ensure my projectors and reactors succeed before persisting my event. 

In this case I would use `persistWithTransaction` instead of `persist` (totally opt-in)

This is a little bit against "pure" event sourcing, but I think it's an improvement of life since if something goes wrong in a projector, it might indicate that something was missed in the aggregate root in terms of validation. 

A simple use case is a unique requirement on the database - if the database throws a unique error - we probably don't want to commit to an event to our event store that violates a unique business rule? Or should the aggregate root, be looking into a projection table to check uniqueness before recording an event? 

Note: Any projectors / reactors that are marked as `ShouldQueue` won't be affected due to their async nature (which I think is a good thing as an examle - a SendWelcomeMail reactor shouldn't stop someone from signing up)

### DIFF
--- a/tests/TestClasses/AggregateRoots/Projectors/ProjectorThatThrowsAnException.php
+++ b/tests/TestClasses/AggregateRoots/Projectors/ProjectorThatThrowsAnException.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\Projectors;
+
+use Exception;
+use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\StorableEvents\MoneyAdded;
+
+class ProjectorThatThrowsAnException extends AccountProjector
+{
+    public function onMoneyAdded(MoneyAdded $event, string $aggregateUuid)
+    {
+        throw new Exception('Computer says no.');
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/TestClasses/AggregateRoots/Projectors/ProjectorThatThrowsAnException.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\Projectors;
+
+use Exception;
+use Spatie\EventSourcing\Tests\TestClasses\AggregateRoots\StorableEvents\MoneyAdded;
+
+class ProjectorThatThrowsAnException extends AccountProjector
+{
+    public function onMoneyAdded(MoneyAdded $event, string $aggregateUuid)
+    {
+        throw new Exception('Computer says no.');
```
freekmurze, line 12: ❤️

brendt:
Indeed, this goes against ES best practices, but I understand why some people would need it.

Just a question though, wouldn't the same work outside the AR?

```php
DB::transaction(function () use ($aggregateRoot) {
    $aggregateRoot->persist();
});
```

If this works I see no compelling reason to add functionality that promotes an anti pattern. If people really want to do it they can so by manually using a transaction.

morrislaptop:
You’re correct 😆 

I do like the syntax of a ‘persistWithTransaction’ method being called on the aggregate but I can easily do this in user land.

brendt:
I think we're going to leave it like that then :) I appreciate the effort though, thanks for all the contributions!

PR #177, opened 2020-12-06
Improve assertions

peter279k:
# Changed log

- Using the `aasertSame` to assert expected is `0` strictly.
- Using the `assertInstanceOf` to assert expected is same as result instance.


freekmurze:
Thanks!

PR #182, opened 2020-12-15
Allow retrieving an aggregate at a specific version

morrislaptop:
I ran into a situation where I needed to retrieve an aggregate for a specific state in the past. This PR adds an optional 2nd argument to the Aggregate Root, so it's possible to get a previous state of an aggregate, instead of just the latest. 

A concrete use-case is when trying to compare a previous state to the latest state, exactly like this PR 🥁 

I also believe this is another solid use case to use aggregates, in addition to the other benefit of making decisions based on past events. 

Here's how it would work with this PR in place:

```php
$account = AccountAggregate::retrieve($uuid, 3);
return $account->balance;
```

Since this is an optional argument, there should be no breaking changes. I also added the argument to `AggregateRoot::loadUuid`, for consistency.

Let me know what you think and if the PR needs any tweaks.

Co-Author: @lukecurtis93
Credit: Quality checked against https://github.com/spatie/simple-excel/pull/43 due to https://twitter.com/freekmurze/status/1337047675611582464

### DIFF
--- a/docs/using-aggregates/creating-and-configuring-aggregates.md
+++ b/docs/using-aggregates/creating-and-configuring-aggregates.md
@@ -55,7 +55,7 @@ This will cause all events with the given `uuid` to be retrieved and fed to the 
 The aggregate also allows you to travel back in time, by passing a version number to retrieve. To check our account balance in the past:
 
 ```php
-$account = MyAggregate::retrieve($uuid, 3);
+$account = AccountAggregate::retrieve($uuid, 3);
 return $account->balance;
 ```
 
--- a/docs/using-aggregates/creating-and-configuring-aggregates.md
+++ b/docs/using-aggregates/creating-and-configuring-aggregates.md
@@ -40,7 +40,7 @@ $myAggregate = new MyAggregate();
 $myAggregate->loadUuid($uuid);
 ```
 
-The `load` method is handy when injecting aggretate roots in constructors or classes where method injection is supported.
+The `load` method is handy when injecting aggregate roots in constructors or classes where method injection is supported.
 
 ```php
 public function handle(MyAggregate $aggregate) {
@@ -52,6 +52,12 @@ public function handle(MyAggregate $aggregate) {
 
 This will cause all events with the given `uuid` to be retrieved and fed to the aggregate. For example, an event `MoneyAdded` will be passed to the `applyMoneyAdded` method on the aggregate if such a method exists.
 
+The aggregate also allows you to travel back in time, by passing a version number to retrieve. To check our account balance in the past:
+
+```php
+$account = MyAggregate::retrieve($uuid, 3);
+return $account->balance;
+```
 
 ## Recording events
 
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -167,7 +167,7 @@ abstract class AggregateRoot
         }
 
         $storedEventRepository->retrieveAllAfterVersion($this->aggregateVersion, $this->uuid)
-            ->takeWhile(function (StoredEvent $storedEvent) use ($maxAggregateVersion) {
+            ->filter(function (StoredEvent $storedEvent) use ($maxAggregateVersion) {
                 return ! $maxAggregateVersion || $storedEvent->aggregate_version <= $maxAggregateVersion;
             })
             ->each(function (StoredEvent $storedEvent) {
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -34,16 +34,7 @@ abstract class AggregateRoot
      *
      * @return static
      */
-    public static function retrieve(string $uuid)
-    {
-        $aggregateRoot = app(static::class);
-
-        $aggregateRoot->uuid = $uuid;
-
-        return $aggregateRoot->reconstituteFromEvents();
-    }
-
-    public static function retrieveAtVersion(string $uuid, int $aggregateVersion)
+    public static function retrieve(string $uuid, int $aggregateVersion = null)
     {
         $aggregateRoot = app(static::class);
 
@@ -52,11 +43,11 @@ abstract class AggregateRoot
         return $aggregateRoot->reconstituteFromEvents($aggregateVersion);
     }
 
-    public function loadUuid(string $uuid): self
+    public function loadUuid(string $uuid, int $aggregateVersion = null): self
     {
         $this->uuid = $uuid;
 
-        return $this->reconstituteFromEvents();
+        return $this->reconstituteFromEvents($aggregateVersion);
     }
 
     public function uuid(): string
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -43,6 +43,15 @@ abstract class AggregateRoot
         return $aggregateRoot->reconstituteFromEvents();
     }
 
+    public static function retrieveAtVersion(string $uuid, int $aggregateVersion)
+    {
+        $aggregateRoot = app(static::class);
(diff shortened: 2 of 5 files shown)


brendt:
There's a few considerations to make.

First we should be careful not to add too many arguments to `retrieve`. If we're not careful, methods like `retrieve` slowly grow over time and end up with four or five optional arguments. Maybe it's best to have something like `rollback` :

```php
```

brendt:
There's a few considerations to make.

First we should be careful not to add too many arguments to `retrieve`. If we're not careful, methods like `retrieve` slowly grow over time and end up with four or five optional arguments. Maybe it's best to have something like `rollbackTo` or `version` :

```php
$account = AccountAggregate::retrieve($uuid)->rollbackTo(3);
$account = AccountAggregate::retrieve($uuid)->version(3);
```

Also: what happens if you apply new events on a rolled back AR? Should we throw an exception? Should event history split at that point? 

That poses the question though: if we throw an exception, then what use has a rolled back AR? You mustn't read state from it directly, so every operation you'd do with it would fail.

So that brings me to what I suppose is the real solution here: better support for temporal queries ("queries in time"): a way to query events and bundle their result into some kind of "read only" object. The pattern is well known in the CQRS world. In fact, I recently saw a good talk explaining the concept: https://www.youtube.com/watch?v=moRDmcl5_WY&t=1487s&ab_channel=GOTOConferences

brendt:
To further elaborate: I think what would be most useful is to support memory-based projections of some sort. You could of course make a dedicated projection that only replays event from a specific AR up until a version, but that would because very tedious very quickly. 

In-memory projections would look like ARs, but wouldn't issue new events, and instead expose replayed state.

morrislaptop:
Thanks for going over this @brendt, we definitely have the same end goal in mind which is great 👍 

The statement `You mustn't read state from it directly` caught me a little off guard, is there a reason for this? It might couple how my aggregate's state is built into the rest of the application - this is the trade off I'm making to have a consistent interpretation of the events. I like the idea of an in-memory projection that looks like an AR, and I would be wary of duplicating event handling logic within this projection as well as the AR if they do a lot of similar things. 

I definitely agree that an AR which is retrieved at a specific version shouldn't raise new events, and this is what happens currently in this PR as the expected version from stored events doesn't match the AR version. 

In my specific example, let's say I've got a `Contract` AR, there are a bunch of events which build up the state of the contract - ClauseAgreed, StartDateAssigned, InterestAgreed etc.. I also have a `SignedContracts` projection which only wants to store contracts which have been agreed to, so it listens for the `ContractSigned` event. When it receives the event, it needs the built state at that particular version to insert all the details into the database. Possible implementations:

a. Retrieve the aggregate at the version of the ContractSigned event
b. Rely on another projection which builds up the state somewhere else?
c. Call an "in-memory" projection and extract the state from there? 

In the case of option c, would it be bad practice for the AR to use this in-memory projection as well, and make decisions based off that?

Happy to move this to the discussions tab :)

brendt:
> is there a reason for this? 

The reasoning is because of CQRS:  it's in general a good idea to split write- and read models when event sourcing. Keep in mind that "model" doesn't mean "eloquent model", instead it means a "representation of real life business processes in code", it's a "model of the real world". There could very well be several classes representing the same read or write model.

That being said: the AR is considered part of the write model. Think of it as a state machine that's able to determine the next steps in your business process, based on the previous steps taken. The goal of CQRS is to prevent your models from growing too large and doing too many things at once. That's why the general consensus is to not expose the AR's state to the outside: it's only meant to help the AR determine the process flow. Besides precent class overgrowth, CQRS also offers more flexibility: our projections (the read model) don't need to follow the same structure as the internal AR state or the events. I'm sure you've identified this as one of the strengths of an event sourced system.

So, while you _could_ expose the AR state,  CQRS _assumes_ most of the read data will benefit from another structure, so why not keep them separate anyway?

Moving on to your concrete question

> I also have a SignedContracts projection which only wants to store contracts which have been agreed to, so it listens for the ContractSigned event. When it receives the event, it needs the built state at that particular version to insert all the details into the database

Can I assume that we're creating a new signed contract when this event happens? As a simplified example, I imagine you have a `ContractCreated` event and a `ContractSigned` event. When the `ContractSigned` event is triggered you want to actually create the signed contract in the database, but you probably need data from the previous `ContractCreated` event as well to create the projection.

I can see why you might want to retrieve state from the aggregate root: you wouldn't want to copy the same data over and over again in events. I agree that this is a shortcoming of our current implementation: there isn't a convenient way to query the existing events. In fact, we've worked around this problem by copying data across events (using DTOs so that we don't have to repeat fields in the event class), and I know how much of a pain this is.

I'd argue that exposing the AR's state isn't the optimal solution either. What you want to do is get the `ContractCreated` event in your projector when handling `ContractSigned`.

So what I'm working on in v5 is exactly this: a better way to write temporal queries (queries in time). It'd look something like this:

```php
class ContractSignedProjector extends projector
{
    public function __construct(
        private EventRepository $eventRepository
    ) {}

    public function onContractSigned(ContractSigned $contractSigned): void
    {
        $contractCreated = $this->eventRepository
            ->whereAggregateRoot($contractSigned->aggregateRootUuid())
            ->whereType(ContractSigned::class)
            ->first();

        $signedContract = SignedContract::create([
            'name' => $contractCreated->name,
            'signed_at' => $contractSigned->date,
        ]);
    }
}
```

You wouldn't even need an in-memory projection, since the query itself can easily be written in this projection.

What I'd love to hear from you: would this actually solve your problems? Do you think this solution is better, as good or worse as exposing AR state?

morrislaptop:
Hi @brendt , thanks again for your detailed response. Understood on not mixing write models from the read model. I think I like the concept of a InMemory projection the most. 

Yes, the `SignedContractsProjector` we'll be creating a new `SignedContract` read model. 

In my project, contracts can be amended, in which case another `ContractSigned` event would be raised and the `SignedContractsProjector` would **update** the existing `SignedContract`.

However, we also need a `ContractVersions` projection, which essentially saves a snapshot of the contract at each signature and stores the dates the contract was effective from / to. 

These different projections could share this In Memory projection without relying on the AR. Technically the AR could also use this in the state machine, if it wanted to. 

For your example, if we only needed data from one event then I wouldn't bother with the InMemory projection. As the number of events that affects the state grows - `ClauseAgreed`, `StartDateAssigned`, `InterestAgreed` etc, then I would create a InMemory to manage this.

I do think what you've suggested is the right approach though, as with access to the raw events I could easily pass them into the InMemory projection.

```php
class ContractSignedProjector extends projector
{
    public function __construct(
        private EventRepository $eventRepository,
        private InMemoryContractProjection $inMemoryContractProjection,
    ) {}

    public function onContractSigned(ContractSigned $contractSigned): void
    {
        $events = $this->eventRepository
            ->whereAggregateRoot($contractSigned->aggregateRootUuid())
           ->whereAggregateVersion('<=', $contractSigned->aggregateVersion())
            ->get();

        $contract = $this->inMemoryContractProject->reconstitute($events);

        $signedContract = SignedContract::create([
            'name' => $contract->name,
            'signed_at' => $contract->date,
        ]);
    }
}
```

brendt:
One more detail: we've decided to call it `EventQuery` instead of `InMemoryProjection`. The difference between them is that you need to manually instantiate the query when you want to use it, and it doesn't receive new events for the rest of the life cycle. It really is a snapshot at a given point in time. To refresh it you'd need to run it again — much like a normal SQL query. I guess you could use it for the ARs state. Our use cases wouldn't benefit from that since the ARs state is very different to what we're projection. But since the query is read only, there's nothing that could go wrong by using it wherever you want. So I'm going to close this one now, you can already check out the changes in the v5 branch.

morrislaptop:
Thanks @brendt, I'll review v5 👍

PR #183, opened 2020-12-16
V5

brendt:
We can discuss all things v5 in this PR.


brendt:
Still on my todo:

- [ ] Move `aggreate_root_uuid` and `stored_event_uuid` to the stored event model as dedicated columns instead of using meta data
- [ ] Look into versioned events
- [ ] Add better filters on the EventRepository:
    - `$repository->whereDate('>= 2020-01-01')`
    - `$repository->whereEvent(OrderCreated::class, …)`
    - `$repository->get()` to gather the filtered results
- [ ] Add support for in-memory projections (reports)
- [ ] Add support for fictional events in in-memory projections

brendt:
A little more background information about those in memory reports and fictional events. Imagine the following question from a client:

> I want a report that shows all earnings for products A and B within January, 2020.

You could write a projection for this, but if the client wants some kind of interface to play around with the period and product filters, you're stuck with generating an infinite amount of different projections. The better approach is to calculate the results on the fly, what I call "in memory projections".

These classes can listen to whatever kind of event, and are replayed on the fly when constructed — similar to how aggregate roots are retrieved, the difference being that these projectors are allowed to combine events from several entities into one.

It could look like this:

```php
$report = new EarningsForProductAndPeriod(
    Period::make('2020-01-01', '2020-02-01'),
    collect([$this->productA, $this->productB]),
);

$this->assertEquals(70_00, $report->totalPrice);
```

Now imagine another client request:

> What if product A had gotten a different price on 2020-02-01, how much would we have earned between 2020-01-01 and 2020-03-01.

Now we not only need a way to replay events on those in-memory projections, but we also need to be able to insert what I'd call "fictional events" in the stream, before replaying. 

This functionality will allow us to solve very complex time-sensitive queries in a clean way.

brendt:
I'm not sure we need those fictional events though, there might be easier solutions to solve that particular feature request.

morrislaptop:
Looking forward to this - I've put EventQuery's in my current app.

If we were running event query's for a single aggregate, would it make sense to include some snapshot functionality?

brendt:
That might be worth looking into it. Will think about it!

brendt:
One big change is that all event listeners are now registered the same way, here's an excerpt from the changelog:

- All event listeners are now registered in the same way: by looking at an event's type hint. This applies to all:
    - Aggregate root `apply` methods
    - Projection listeners
    - Reactor listeners
    - Event queries

### A note on changed listeners

Since most code is probably already type hinting events, the listener change is likely to not have an impact on your code. It's good to know though that you don't have to worry about certain naming conventions any more:

- In **aggregate roots**, you don't have to prefix apply methods with `apply` anymore if you don't want to
- In **projectors**, you don't need a manual mapping anymore, neither does the event variable need to be called `$event`
- In **reactors**, you don't need a manual mapping anymore, neither does the event variable need to be called `$event`
- **Event queries** are a new concept and work in the same way

Here's an example:

```php
class MyProjector extends Projector
{
    public function anEventHandlerWithAnotherName(MyEvent $eventVariableWithAnotherName): void
    {
        // This handler will automatically handle `MyEvent`
    }
}
```

morrislaptop:
Woah? Closing and deleting?

brendt:
We're going to completely rework this package with regards to the event sourcing course we're making; we'll keep you posted!

freekmurze:
@morrislaptop to stay in the loop, subscribe here: https://spatie.be/event-sourcing A full promotional website is coming soon! Sneak preview:

CleanShot 2021-03-03 at 12 10 58

PR #184, opened 2020-12-17
improved mapping of accepted events

DaanMeijer:
Hey guys, saw you work on this in the stream (not live, or I would have commented), think this is the way to go :)


brendt:
Thanks, it's already fixed in the v5 branch.

PR #185, opened 2020-12-18
V5 extra: multiple methods for the same event

DaanMeijer:
I heard you talking about this on stream, and thought: this is a nice exercise for me to try my hand with tests and all that jazz. I added an extra test and implemented the logic.


brendt:
Thanks, it's already added in the v5 branch :)

PR #189, opened 2021-01-13
Events with arrays not deserialized correctly

morrislaptop:


freekmurze:
@brendt could you take a look at this one?

morrislaptop:
Looking at this as well, I'm currently exploring - https://github.com/symfony/symfony/issues/25524

brendt:
The first thing I'd normally do is create my own simple JsonSerializer and use that one instead of the one provided by the package (using Symfony's underneath).

It looks something like this:

```php
class JsonSerializer implements EventSerializer
{
    public function serialize(ShouldBeStored $event): string
    {
        return json_encode(['event' => serialize($event)]);
    }

    public function deserialize(string $eventClass, string $json, string $metadata = null): ShouldBeStored
    {
        return unserialize(json_decode($json, true)['event']);
    }
}
```

I think it's a matter of coding style: I've gotten used to sanitising my data before constructing an event, so I'm pretty sure I'm always dealing with something that's json serializable. 

That's just some background information if someone were to stumble upon this PR and needed a quick fix.

With regards to the PR: do you know why the problem is happening and how it's fixed?

morrislaptop:
@brendt I think that JSON serializer would only work with scalar types - removing all the benefits of the Symfony serializer.

To specify array types in PHP we have to use a docblock as generics aren't supported. By default the Symfony serializer only reads the strongly typed properties of the class and not docblocks - without a type it just dumps the array into the deserialized object rather than denormalizing it into a collection of objects. The Symfony serializer also needs the ArrayDenormalizer which picks up on the [] suffix to indicate an array.

The solution is to add the ArrayDenormalizer and use a custom ObjectNormalizer in config/event-sourcing.php

```php
'event_normalizers' => [
        Spatie\EventSourcing\Support\CarbonNormalizer::class,
        Spatie\EventSourcing\Support\ModelIdentifierNormalizer::class,
        Symfony\Component\Serializer\Normalizer\DateTimeNormalizer::class,
        Symfony\Component\Serializer\Normalizer\ArrayDenormalizer::class,
        Spatie\EventSourcing\Support\ObjectNormalizer::class,
    ],
```

```
<?php

namespace Spatie\EventSourcing\Support;

use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
use Symfony\Component\Serializer\NameConverter\NameConverterInterface;
use Symfony\Component\Serializer\Mapping\ClassDiscriminatorResolverInterface;
use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactoryInterface;
use Symfony\Component\Serializer\Normalizer\ObjectNormalizer as SymfonyObjectNormalizer;

class ObjectNormalizer extends SymfonyObjectNormalizer
{
    public function __construct(ClassMetadataFactoryInterface $classMetadataFactory = null, NameConverterInterface $nameConverter = null, PropertyAccessorInterface $propertyAccessor = null, PropertyTypeExtractorInterface $propertyTypeExtractor = null, ClassDiscriminatorResolverInterface $classDiscriminatorResolver = null, callable $objectClassResolver = null, array $defaultContext = []) {
        parent::__construct(
            $classMetadataFactory,
            $nameConverter,
            $propertyAccessor,
            $propertyTypeExtractor ?? new PhpDocExtractor(),
            $classDiscriminatorResolver,
            $objectClassResolver,
            $defaultContext
        );
    }
}
```

The above can be done in user land for sure, but might be an expectation of the package that this would work by default.

brendt:
Gotcha! @freekmurze I'm fine adding this, you as well?

FYI the event itself is serialized using PHP's `serialize` method, so it can handle objects as well.

freekmurze:
Yup, I'm ok with this too! 👍

brendt:
Merged and tagged: https://github.com/spatie/laravel-event-sourcing/releases/tag/4.7.1

mbardelmeijer:
This seems to break if `phpdocumentor/reflection-docblock` isn't installed in production. 

Seems to be a breaking change. Perhaps `phpdocumentor/reflection-docblock` should be added to the `require` part of this package?

After updating to 4.7.1, the following exception is thrown in production:
`Unable to use the "Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor" class as the "phpdocumentor/reflection-docblock" package is not installed.`

brendt:
@mbardelmeijer very sorry for the inconvenience. 4.7.2 should fix it: https://github.com/spatie/laravel-event-sourcing/releases/tag/4.7.2

PR #192, opened 2021-01-27
Add event versioning proof of concept

brendt:
This branch is a proof of concept for adding first class event versioning support.

The idea is as follows:

- Every time an event is stored, a version is kept in the database, by default this version will be `1`
- Whenever an event is restored from the database, the developer has the option to hook into the deserialization process to check the event version. If it's an older version they can make on-the-fly changes to the original data so that it works with the new event class.

A simple example: we've got an event `VersionedEvent` that used to have a field called `name`. Business rules changed over time though, and now this event should only contain a field called `uuid`. The business also decided that, if there's no UUID present for old events, the name should be used as its "UUID".

Previously, the event looked like this:

```php
class VersionedEvent extends ShouldBeStored
{
    public function __construct(
        public string $name
    ) {
    }
}
```

However, as of the business changes, the event now looks like this:

```php
class VersionedEvent extends ShouldBeStored
{
    public function __construct(
        public string $uuid
    ) {
    }
}
```

So we need a way to:

- indicate that this is version two of the same event; and to
- be able to unserialize the v1 data into the new version

The first requirement is straight forward: we can use the new `#[EventVersion]` attribute:

```php
#[
    EventVersion(2),
]
class VersionedEvent extends ShouldBeStored
{
    public function __construct(
        public string $uuid
    ) {
    }
}
```

Whenever a new `VersionedEvent` is saved into the database, it will now be known with its version 2.

Next we still need a way to unserialize the old data. To do this I decided to hook into the components that were already present in the package and allow users to configure their own serializer on a per-event basis, again using an attribute:

```php
#[
    EventVersion(2),
    EventSerializerAttribute(VersionedEventSerializer::class),
]
class VersionedEvent extends ShouldBeStored
{
    public function __construct(
        public string $uuid
    ) {
    }
}
```

The `VersionedEventSerializer` looks like this:

```php
class VersionedEventSerializer extends JsonEventSerializer
{
    public function deserialize(
        string $eventClass,
        string $json,
        int $version,
        string $metadata = null
    ): VersionedEvent {
        $data = json_decode($json, true);

        if ($version === 1) {
            $data['uuid'] = $data['name'] ?? '';
            unset($data['name']);
        }

        return new VersionedEvent(...$data);
    }
}
```

As you can see, the version is now provided withen the `deserialize` and we can do checks on it. 

A few notes with this PR:

- The method signature of `\Spatie\EventSourcing\EventSerializers\EventSerializer::deserialize` has changed, adding the `$version` variable
- The `stored_events` table needs a new column called `event_version`
- There's no hooks yet to use custom serializers for when you're serializing an event. Even though I can't come up with any realistic use cases, I think it's best to also support it, given that the `EventSerializerAttribute(VersionedEventSerializer::class),` attribute indicates both serialization and deserialization. An alternative could be to step away from the existing serializers and add a layer on top. I think that's an unnecessarily complex solution. Though we can discuss it.


brendt:
BTW: This PR is targeted at the v5 branch

rubenvanassche:
```php
#[
    EventVersion(2),
    EventSerializerAttribute(VersionedEventSerializer::class),
]
class VersionedEvent extends ShouldBeStored
{
    public function __construct(
        public string $uuid
    ) {
    }
}
```

Couldn't we rename `EventSerializerAttribute` to `EventSerializer` seems a a bit more fitting with `EventVersion`?

```php
class VersionedEventSerializer extends JsonEventSerializer
{
    public function deserialize(
        string $eventClass,
        string $json,
        int $version,
        string $metadata = null
    ): VersionedEvent {
        $data = json_decode($json, true);

        if ($version === 1) {
            $data['uuid'] = $data['name'] ?? '';
            unset($data['name']);
        }

        return new VersionedEvent(...$data);
    }
}
```

Would we ever need a json representation of the event? I think we always will use that `$data` variable which is decoded json. Maybe pass the json directly as an array through the function? Since every serializer probably will have that first statement?

brendt:
> Couldn't we rename EventSerializerAttribute to EventSerializer seems a a bit more fitting with EventVersion?

That's the case, but I had to import it as an alias in my test case because of inline classes :) This wouldn't happen IRL: https://github.com/spatie/laravel-event-sourcing/pull/192/files#diff-2e1ebcf697039e71cd0c7462606d9a872492a7d9b7ce0459de720c69b3809f99

> Would we ever need a json representation of the event? I think we always will use that $data variable which is decoded json. Maybe pass the json directly as an array through the function? Since every serializer probably will have that first statement?

The problem here is that I'm building on top of the existing serializers, so I want to avoid unnecessary breaking changes if possible.

PR #196, opened 2021-01-29
Entities

brendt:
TODO:

- [x] Tests
- [ ] Docs

### DIFF
--- a/src/AggregateRoots/AggregateEntity.php
+++ b/src/AggregateRoots/AggregateEntity.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Spatie\EventSourcing\AggregateRoots;
+
+use Spatie\EventSourcing\EventHandlers\AppliesEvents;
+use Spatie\EventSourcing\StoredEvents\ShouldBeStored;
+use Spatie\EventSourcing\StoredEvents\StoredEvent;
+
+abstract class AggregateEntity
+{
+    use AppliesEvents;
+
+    public function __construct(
+        private AggregateRoot $aggregateRoot
+    ) {
+    }
+
+    protected function aggregateRootUuid(): string
+    {
+        return $this->aggregateRoot->uuid();
+    }
+
+    protected function recordThat(ShouldBeStored $event): void
+    {
+        $this->aggregateRoot->recordThat($event);
+    }
+
+    public function apply(StoredEvent|ShouldBeStored ...$storedEvents): void
+    {
+        foreach ($storedEvents as $storedEvent) {
+            $this->applyStoredEvent($storedEvent);
+        }
+    }
+}
--- a/src/AggregateRoots/AggregateEntity.php
+++ b/src/AggregateRoots/AggregateEntity.php
@@ -25,7 +25,7 @@ abstract class AggregateEntity
         $this->aggregateRoot->recordThat($event);
     }
 
-    public function apply(StoredEvent|ShouldBeStored ...$storedEvents): void
+    public function apply(StoredEvent | ShouldBeStored ...$storedEvents): void
     {
         foreach ($storedEvents as $storedEvent) {
             $this->applyStoredEvent($storedEvent);
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -36,7 +36,7 @@ abstract class AggregateRoot
 
     public function __construct()
     {
-        collect((new ReflectionClass($this))->getProperties(ReflectionProperty::IS_PROTECTED|ReflectionProperty::IS_PUBLIC))
+        collect((new ReflectionClass($this))->getProperties(ReflectionProperty::IS_PROTECTED | ReflectionProperty::IS_PUBLIC))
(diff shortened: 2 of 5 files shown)


brendt:
This PR adds support for entities as discussed here: https://github.com/spatie/laravel-event-sourcing/discussions/195

tests/AggregateRoots/AggregateEntityTest.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $cart::retrieve(self::CART_UUID);
+
+        $this->assertExceptionThrown(function () use ($cart) {
+            $cart->checkout();
+        });
+    }
+}
+
+class Cart extends AggregateRoot
```
freekmurze, line 48: Could you move these test classes each in their own files in a `TestClasses` namespace? That seems cleaner to me.

src/AggregateRoots/AggregateRoot.php
```diff
@@ -29,6 +30,35 @@ abstract class AggregateRoot
...
     protected static bool $allowConcurrency = false;
 
+    /** @var \Spatie\EventSourcing\AggregateRoots\AggregateEntity[] */
+    private array $entities = [];
+
+    public function __construct()
+    {
+        $reflectionClass = new ReflectionClass($this);
+
+        foreach ($reflectionClass->getProperties() as $property) {
```
freekmurze, line 40: Could we refactor this to collections? I think that will be nicer as we could use a few `filter` calls and at the end have a `each` where we could use a `AggregateEntity` in the callable.

riasvdv:
In this implementation, how would the Entity access state from the AggregateRoot? https://github.com/spatie/laravel-event-sourcing/discussions/195#discussioncomment-317440

src/AggregateRoots/AggregateEntity.php
```diff
@@ -0,0 +1,34 @@
...
+use Spatie\EventSourcing\EventHandlers\AppliesEvents;
+use Spatie\EventSourcing\StoredEvents\ShouldBeStored;
+use Spatie\EventSourcing\StoredEvents\StoredEvent;
+
+abstract class AggregateEntity
+{
+    use AppliesEvents;
+
+    public function __construct(
+        private AggregateRoot $aggregateRoot
```
freekmurze, line 14: I'm thinking we could use `protected` here

src/AggregateRoots/AggregateEntity.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return $this->aggregateRoot->uuid();
+    }
+
+    protected function recordThat(ShouldBeStored $event): void
+    {
+        $this->aggregateRoot->recordThat($event);
+    }
+
+    public function apply(StoredEvent | ShouldBeStored ...$storedEvents): void
```
freekmurze, line 28: ❤️  union types

src/AggregateRoots/AggregateRoot.php
```diff
@@ -29,6 +30,35 @@ abstract class AggregateRoot
 
     protected static bool $allowConcurrency = false;
 
+    /** @var \Spatie\EventSourcing\AggregateRoots\AggregateEntity[] */
+    private array $entities = [];
+
+    public function __construct()
```
freekmurze, line 36: Pretty cool that we can automatically new up these entities 👍

freekmurze (changes requested):
Very nice work!

brendt:
@riasvdv The AR can either pass state to its entities when calling their methods, or the AR could expose its state via getters. There's unfortunately no concept op friend classes in PHP so such getters would be public. It's the responsibility of the developer to not use those getters outside the aggregate.

src/AggregateRoots/AggregateRoot.php
brendt: Seemed like the right thing to do. Unfortunately it also means the entity properties should be protected. If you want to keep them private you should manually set them up, but I wouldn't advice to do that.

src/AggregateRoots/AggregateEntity.php
brendt: Great idea!

src/AggregateRoots/AggregateRoot.php
brendt: I agree. Even better would be if we made a package that integrates reflection with collections and add other useful stuff. One that I can think of is `$type->accepts($variable)` instead of manually comparing types. Anyways, that another discussion and I'll refactor it for now.

src/AggregateRoots/AggregateEntity.php
brendt: Just tried it, unfortunately we can't do it because an AR is resolved via the container, so the constructor must be public.

src/AggregateRoots/AggregateRoot.php
freekmurze: maybe we could add a method `ensure` on our macro collections package?

It should throw an exception if there's an item in the collection that isn't of any of the types passed to `ensure`. 

Might also be PR worthy to Laravel.

PR #209, opened 2021-04-17
Update writing-your-first-aggregate.md

guifelix:
removing duplicate method name & replacing with correct one


freekmurze:
Thanks!

PR #210, opened 2021-04-19
Add drop table for snapshot and stored_event to the migration stub

sawirricardo:
This PR adds dropIfExists for snapshot and stored_event table to the migration stub.


freekmurze:
There are too many changes in this PR. Please try again with a new PR.

PR #211, opened 2021-04-20
Update writing-your-first-projector.md

WyattCast44:
Greeting Spatie Team 👋

I added a short section on how to use existing model factories with the given example. I was working on my first application using this package and I could not find any example of how to do this - this is the solution I came up with. Please feel free to edit the wording or expand/revise as needed. If you have another method I would love to learn about it. 

Cheers Wyatt 🍻


freekmurze:
Thank you!

PR #216, opened 2021-05-07
Fix TypeError in EloquentStoredEvent Model

MrSnoop:
This should fix a typeError
 Type of Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEvent::$guarded must be array (as in class Illuminate\Database\Eloquent\Model)


freekmurze:
We can't add this because that typehint isn't set on the base class.

PR #224, opened 2021-06-10
fix migrations stubs location

khanalpride:
Looks like migration stubs are on different location.


freekmurze:
Seems like the tests are failing for this one.

khanalpride:
@freekmurze 
I just moved the database migration stubs to the location expected by `spatie/laravel-package-tools`

And when I did that, the `php artisan vendor:publish` command run successfully.

I don't think the test have to do anything with that change.

khanalpride:
@freekmurze ah I found the issue.

freekmurze:
Thank you!

PR #237, opened 2021-07-22
small suggestion: jsonb columns instead of json

morrislaptop:
`jsonb` is generally considered [more performant](https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/) than `json` 

on db's that don't support `jsonb`, Laravel will automatically fallback to `json`


brendt:
Ah thanks!

brendt:
Tagged: https://github.com/spatie/laravel-event-sourcing/releases/tag/5.0.5

PR #243, opened 2021-08-04
docs: update 'Models with timestamps' section

keizah7:
```
public function onAccountCreated(StoredEvent $storedEvent, AccountCreated $event) {
        Account::create(array_merge($event->accountAttributes, ['created_at' => $storedEvent->created_at, 'updated_at' => $storedEvent->created_at]));
}
```

I was trying to find out how docs writer [getting](https://spatie.be/docs/laravel-event-sourcing/v5/advanced-usage/replaying-events#models-with-timestamps) `$storedEvent` instance in event arguments and found a way save code line using ShouldBeStored method. It looks cleaner.

Also if you are planning of accepting this pull-request. Perhaps we should add the functionality to return `createdAt` value using `app.timezone` config property?


brendt:
Thanks!

PR #248, opened 2021-09-08
V6

brendt:
In order to support catch-all event handlers, we'll need to make a few changes to the `EventHandler` interface

```diff
-    public function handles(): array;
+    public function handles(StoredEvent $storedEvent): bool;

-    public function handle(StoredEvent $event);
+    public function handle(StoredEvent $storedEvent): void;
```

I think the feature is good enough to warrant a new major. 

For context: we'll now use spatie/better-types to discover event handlers, this means that event listeners can now also listen to interfaces, unions and even `object`:

```php
class CatchAllReactor extends Reactor
{
    public function __invoke(ShouldBeStored $event): void
    {
    }
}

class CatchAllReactor extends Reactor
{
    public function __invoke(object $event): void
    {
    }
}

class CatchAllReactor extends Reactor
{
    public function __invoke(InterfaceA|InterfaceB $event): void
    {
    }
}
```


brendt:
Todo:

- [ ] Integrate `Handlers` in ARs
- [ ] Remove project `Handlers` implementation
- [ ] Fix failing replay test

composer.json
```diff
@@ -29,6 +29,7 @@
         "illuminate/support": "^8.0",
         "league/flysystem": "^1.1.3",
         "phpdocumentor/reflection-docblock": "^5.2",
+        "spatie/better-types": "dev-main",
```
abenerd, line 32: I get that the `dev-main` is needed for this to work now, but would we wait for `v1` to be released or would we use the latest version that is available when this PR is ready? Personally I'd say we wait for the `v1` release of `spatie/better-types`.

brendt: Yeah I'm going to release v1 of `better-types` in preparation of this release :)

abenerd: Ah, this is great.
And great work here, btw!

PR #249, opened 2021-09-10
Add backwards compatibility for eventId

Robertbaelde:
This PR introduces backwards compatibility for events that don't have the eventId in their metadata. 

Imho eventId shouldn't be in the metadata in the first place. This would also prevent saving, and immediately updating in the EloquentStoredEventRepository. Curious what the thoughts behind this are?

### DIFF
--- a/src/StoredEvents/StoredEvent.php
+++ b/src/StoredEvents/StoredEvent.php
@@ -142,6 +142,7 @@ class StoredEvent implements Arrayable
             );
 
             $this->event->setMetaData(optional($this->meta_data)->toArray());
+            $this->event->setStoredEventId($this->id);
         } catch (Exception $exception) {
             throw InvalidStoredEvent::couldNotUnserializeEvent($this, $exception);
         }
--- a/tests/Models/StoredEventTest.php
+++ b/tests/Models/StoredEventTest.php
@@ -3,6 +3,7 @@
 namespace Spatie\EventSourcing\Tests\Models;
 
 use Carbon\Carbon;
+use Spatie\EventSourcing\Enums\MetaData;
 use Spatie\EventSourcing\Facades\Projectionist;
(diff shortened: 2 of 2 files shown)


Robertbaelde:
@freekmurze don't want to push, but is there anything I can do to help this getting merged?

freekmurze:
@Robertbaelde forgot about this one, my apologies
@brendt could you take a look at this one?

brendt:
Could you elaborate what problem this solves?

Robertbaelde:
@brendt events stored by pre v5 versions of this package, don't have the storedEventId in the metadata field. This causes replays to fail in this version. Since it uses the metadata to get the stored eventId, instead of the id from the stored event

This PR sets the stored event id based on the ID of the stored event, not based on the field in the metadata. 

Really curious why we actually need the eventId in the metadata column? Would be great if you could elaborate on that. Since it comes at the costs of writing to the database twice (save & update afterwards) and saving data in the model twice.

Robertbaelde:
@brendt Do you need more info from me? :)

Robertbaelde:
@brendt the fact that the migration is not necessary. I rather not update the event stream. 

I really don't understand why we'd need the stored event ID in the metadata? 

I get is (sort of) to put it in the storedEvent objects metadata. But why do we need it in the database there as well? It saves a line of code on the read side, but costs us an extra db write (and extra loc) on the write side.

brendt:
If I'm not mistaken, it's because we wanted to expose the ID without introducing a new breaking change back in v4. I agree that it could be implemented differently. In fact, I plan on doing more major refactors in the future, but it's not on my short-term schedule.

I'm going to close this PR: v5 isn't meant to be compatible with previous versions, and I don't want to maintain extra code and tests just for backwards compatibility. The right way to go is by migrating your event stream; it's a valid approach described by Greg Young’s book called "Versioning in an Event Sourced System" — in case you'd like to do some more reading on the topic.

PR #251, opened 2021-09-10
Fix AggregateRoot return types for static analysis

erikgaal:
When using a static analysis tool such as PHPStan, a type error is returned when using the following code snippet.

```php
namespace App\Aggregates;

use Spatie\EventSourcing\AggregateRoots\AggregateRoot;

class AccountAggregate extends AggregateRoot
{
    public function createAccount(string $name, string $userId): self
    {
        return $this->recordThat(new AccountCreated($name, $userId));
    }
}
```

```
Method App\Aggregates\AccountAggregate::add() should return App\Aggregates\AccountAggregate but returns Spatie\EventSourcing\AggregateRoots\AggregateRoot.
```

The problem is with the _short_ notation directly returning the result of the `recordThat` call. The return type resolves to `Spatie\EventSourcing\AggregateRoots\AggregateRoot`, instead of the current class due to the `self` return type. However, if we use the `static` return type, it is resolved correctly to the current class.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -54,7 +54,7 @@ abstract class AggregateRoot
         return $aggregateRoot->reconstituteFromEvents();
     }
 
-    public function loadUuid(string $uuid): self
+    public function loadUuid(string $uuid): static
     {
         $this->uuid = $uuid;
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you!

PR #269, opened 2021-10-02
Fix urls pointing to previous version

quintenbuis:
Some urls still point to v4 which can cause confusion. This PR fixes that.

PS. Can the `hacktoberfest-accepted` label be added on the PR? 🥳


freekmurze:
Thanks!

PR #273, opened 2021-10-07
[Docs] Add `EloquentStoredEvent` import to example

stevebauman:
I wasn't able to easily locate what the class' namespace the `EloquentStoredEvent` model had in the docs, so I figured it may help others out as well if we add the full `use` import.

No hard feelings on closure ❤️ , thanks for your hard work!


freekmurze:
Thanks!

PR #279, opened 2021-11-07
[Docs] Fix wrong operator for onMoneySubtracted

avosalmon:
The value should be decremented on the `MoneySubtracted` event.


freekmurze:
Thank you!

PR #284, opened 2021-11-16
Changed cursor() into lazyById() to preserve memory when working with large amount of events

etahamer:
In one of my projects I have several millions of events and when I tried to replay them all, my process was dying because of being out of memory. I have solved this issue by changing cursor() into lazyById()


freekmurze:
Thanks!

PR #286, opened 2021-11-16
Fixed tests/VersionedEventTest.php::a_versioned_event_can_be_restored 

etahamer:
Phpunit tests failed in `a_versioned_event_can_be_restored` method of `tests/VersionedEventTest.php` due to 'uuid' property not being passed. As I have seen, event was being given `name` instead of `uuid` argument with correct value, so I have simply renamed property and test passes once again.


freekmurze:
Thank you, much appreciated

PR #289, opened 2021-11-17
Added check whether lazyById method is present in query builder. 

etahamer:
…tigate problem when lower laravel/framework versions than v8.34.0 is used.


freekmurze:
Instead of adding this, please update the constraints in `composer.json`

PR #290, opened 2021-11-17
Set minimum version of illuminate/database to ^8.34

etahamer:
This change is required because lazyById method is available from illuminate/database 8.34 version.


freekmurze:
Thank you!

Issue #276, opened 2021-10-23
Persisting events while testing aggregate


maciejfikus:
Hi guys,
I noticed that after running my aggregate test suits all asserted events are actually getting persisted to event store. I believe that it is some sort of bug, isn't it? Otherwise, how am I supposed to test them without actually using database (even litesql one).
It is done here: https://github.com/spatie/laravel-event-sourcing/blob/main/src/AggregateRoots/FakeAggregateRoot.php#L33

How to reproduce:
`Aggregate::fake($uuid)->given([new SomeEvent('data')]);`

PR #292, opened 2021-11-25
Aggregate tests without db

morrislaptop:
This issue solves #276 

I assume the `persist` was used to clear the events so that we could easily see what events were recorded during the `when` stage. 

The [first POC](https://github.com/spatie/laravel-event-sourcing/commit/d6814bcd6ef63ffc573cf8350c1bbce766620dd2) of this was to simply call the `getAndClearRecordedEvents` on the AggregateRoot. 

However, we can achieve the same result by simply [remembering how many `given` events there were](https://github.com/morrislaptop/laravel-event-sourcing/blob/56fdc8f47a7ca525eafc632839db9a9c9f3f6414/src/AggregateRoots/FakeAggregateRoot.php#L34) and [removing them](https://github.com/morrislaptop/laravel-event-sourcing/blob/56fdc8f47a7ca525eafc632839db9a9c9f3f6414/src/AggregateRoots/FakeAggregateRoot.php#L174) when doing our assertions.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -201,7 +201,7 @@ abstract class AggregateRoot
         }
     }
 
-    public function getAndClearRecordedEvents(): array
+    protected function getAndClearRecordedEvents(): array
     {
         $recordedEvents = $this->recordedEvents;
 
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -201,7 +201,7 @@ abstract class AggregateRoot
         }
     }
 
-    protected function getAndClearRecordedEvents(): array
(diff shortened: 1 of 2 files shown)


brendt:
Thanks, makes sense!

brendt:
Released here: https://github.com/spatie/laravel-event-sourcing/releases/tag/6.0.1

Robertbaelde:
Nice work @morrislaptop 

Be aware, I think it doesn't fully fixes the referenced issue though. Since calling the static fake method on an aggregate still first retrieves it from the the db before instantiating the FakeAggregateRoot. This still requires a db, which you ideally won't have instantiated in your unit tests. 

(https://github.com/spatie/laravel-event-sourcing/blob/a8ee599d47361156ed11a18b46b55bd3b922b79f/src/AggregateRoots/AggregateRoot.php#L288)

PR #294, opened 2021-11-26
skip retrieve on aggregate::fake

morrislaptop:
Follow up on #292 from https://github.com/spatie/laravel-event-sourcing/pull/292#issuecomment-979821746

If we are calling the fake method can we assume there'll be no prior events and skip the `retrieve` method.


brendt:
Thanks!

brendt:
Tagged: https://github.com/spatie/laravel-event-sourcing/releases/tag/6.0.2

PR #295, opened 2021-11-28
downgrade symfony finder

morrislaptop:
https://github.com/spatie/laravel-event-sourcing/blob/eb7c3ee5ac741544c2f81083b727ff966f617e2b/composer.json#L35

5.3.8 doesn't seem to exist. All tests pass on 5.3.7. 🤷


freekmurze:
Thanks!

brendt:
Thanks! So strange… I'm sure I tested this. You're absolutely right though and I most likely missed something or mistyped somewhere, thanks for letting us know and fixing it!

PR #296, opened 2021-12-03
Don't mutate original events in `FakeAggregateRoot::getRecordedEventsWithoutUuid`

inxilpro:
When you use something like `assertRecorded` on an aggregate root fake, it calls `getRecordedEventsWithoutUuid`:

https://github.com/spatie/laravel-event-sourcing/blob/01b17e0ba6d1795acf7b91db5475b7a0a36ea4ac/src/AggregateRoots/FakeAggregateRoot.php#L163-L175

This method removes any generated metadata from the event so that it can be safely compared, but it mutates the original event instead of cloning it or creating a new copy. This means that if you then want to use the events on the aggregate root later in your test, those metadata values are gone.

This PR just clones the event before mutating it so that if you need the generated metadata later in your test it is still available.

### DIFF
--- a/src/AggregateRoots/FakeAggregateRoot.php
+++ b/src/AggregateRoots/FakeAggregateRoot.php
@@ -122,6 +122,7 @@ class FakeAggregateRoot
         $expectedEvents = Arr::wrap($expectedEvents);
 
         $appliedEvents = array_map(function (ShouldBeStored $event) {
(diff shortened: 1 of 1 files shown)


brendt:
Thanks!

brendt:
Tagged: https://github.com/spatie/laravel-event-sourcing/releases/tag/6.0.4

PR #298, opened 2021-12-16
Exclude given events in FakeAggregateRoot::assertNotRecorded

bastien-phi:
In the documentation about testing aggregate (https://spatie.be/docs/laravel-event-sourcing/v5/using-aggregates/testing-aggregates), there is this example 

```php
/** @test */
public function it_will_not_make_subtractions_that_would_go_below_the_account_limit()
{
    AccountAggregateRoot::fake()
        ->given([new AccountCreated('Luke'), new MoneySubtracted(4999)])
        ->when(function (AccountAggregate $accountAggregate) {
            $accountAggregate->subtractMoney(2);
        })
        ->assertRecorded(new AccountLimitHit(2))
        ->assertNotRecorded(MoneySubtracted::class);
}
```

Actually this test fails to assert that `MoneySubstracted` was not recorded because it is given in the FakeAggregateRoot.

This PR fixes it.

### DIFF
--- a/src/AggregateRoots/FakeAggregateRoot.php
+++ b/src/AggregateRoots/FakeAggregateRoot.php
@@ -85,7 +85,7 @@ class FakeAggregateRoot
 
     public function assertNotRecorded($unexpectedEventClasses): self
     {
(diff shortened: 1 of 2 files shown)


brendt:
Unfortunately we wouldn't be able to merge this in without potentially breaking everyone's testsuite. We should change the docs instead.

PR #304, opened 2021-12-28
Repaired URLs to current documentation version.

Mul-tiMedia:
Needed some documentation, but got a `This site can’t be reached` on https://docs.spatie.be.


freekmurze:
Thanks!

PR #305, opened 2021-12-29
Add line break to increase readability

craigpotter:
Add a line break so use statement and call to static method are on separate lines


freekmurze:
Thanks!

PR #306, opened 2022-01-05
Change event_version column type from int to unsigned tinyint

rapkis:
Changed the `event_version` column type from int to unsigned tinyint in the `stored_events` table migration stub.

The event version shouldn't be negative, therefore, the new column type would be more semantically correct for the use-case of this feature

### DIFF
--- a/database/migrations/create_stored_events_table.php.stub
+++ b/database/migrations/create_stored_events_table.php.stub
@@ -12,7 +12,7 @@ class CreateStoredEventsTable extends Migration
             $table->bigIncrements('id');
             $table->uuid('aggregate_uuid')->nullable();
             $table->unsignedBigInteger('aggregate_version')->nullable();
-            $table->integer('event_version')->default(1);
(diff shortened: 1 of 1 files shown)


brendt:
While I don't mind the change, I find it hard to understand the justification for it: the event version column is used to store the version when working with versioned events, that number should only increase when changes to events were made, and could realistically not be higher then maybe 10 or 20 versions (although that's already _a lot_)

rapkis:
You're totally right. Not sure why I didn't mention it in the original description, so let me add it here: 

Correct me if I'm wrong, but the event version _shouldn't_ be negative, therefore, the new column type would be more semantically correct for use-case of the feature.

That's my main reasoning for submitting this pull request.

brendt:
maybe `tinyint` would be more appropriate then?

rapkis:
Good point, thanks for the feedback. I've applied the changes

brendt:
Thanks!

brendt:
Tagged: 6.0.5

Issue #307, opened 2022-01-08
Aggregate Partial is ignored when creating a snapshot


saschanowak:
I'm playing around with this package and noticed, that when following the documentation for [aggregate partials](https://spatie.be/docs/laravel-event-sourcing/v5/advanced-usage/aggregate-partials) and creating a snapshot of the aggregate root, the state of the partial is not stored in the snapshot. 
Not sure if that is intended or a bug, as I didn't find a test for this usecase and the documentation didn't mentions snapshots and aggregate partials.

etahamer:
@nlx-sascha

> I'm playing around with this package and noticed, that when following the documentation for [aggregate partials](https://spatie.be/docs/laravel-event-sourcing/v5/advanced-usage/aggregate-partials) and creating a snapshot of the aggregate root, the state of the partial is not stored in the snapshot. Not sure if that is intended or a bug, as I didn't find a test for this usecase and the documentation didn't mentions snapshots and aggregate partials.

1. Is your partial aggregate root is Public property of your aggregate?
2. Are the properties that you expect to snapshot public in aggregate partial?

nlx-sascha:
Hey @etahamer,

thanks for the quick response.

Without trying it deeper I think making the AggregatePartial public does not feel right. Correct me if I'm wrong, but the interaction with the AggregatePartial should not be possible from outside. Thats when you need to proxy all functions from the AggregateRoot to the AggregatePartial like in the documentation example.
Also when reading the code the `useState` function only handles scalar values and does not convert the values of the aggregate partial to the correct object:
![image](https://user-images.githubusercontent.com/1643495/148660770-aa67b266-1295-448b-8679-f26a841071a1.png)

brendt:
Partials should indeed not be made public. I'd say this issue is mostly an oversight, though I'm not sure whether it can be easily fixed. I welcome a PR that addresses the issue.

PR #309, opened 2022-01-11
Snapshotting of aggregate partial

nlx-sascha:
Only public properties are includes in snapshots currently.
This change adds the feature to automatically include all
aggregate partials in the snapshot. Also added a small
feature that allows to also include internal state in the
snapshot by adding an attribute to the property that should
be included.

Resolves: #307

### DIFF
--- a/src/Attributes/IncludeInSnapshot.php
+++ b/src/Attributes/IncludeInSnapshot.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Spatie\EventSourcing\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class IncludeInSnapshot
+{
+}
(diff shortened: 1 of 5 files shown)


src/Attributes/IncludeInSnapshot.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Spatie\EventSourcing\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class IncludeInSnapshot
```
nlx-sascha, line 8: I'm not 100% happy with the name and really open to better suggestions.

brendt:
I'm wondering what will happen when people have snapshots stored in an older version, then update, and then try to restore that snapshot with these changes — with and without aggregate partials. 

Have you done some testing in that project you mentioned in the original issue?

freekmurze:
Closing because of inactivity.

PR #310, opened 2022-01-13
Laravel 9.x Compatibility

erikgaal:
The `league/flyststem` requirement had to be widened to include `^2.0` as well. However, I noticed that we're not using any code from that package explicitly anyway, so I removed it.

EDIT:

I accidentally based this PR on an older version of the package. I see that it's blocked by https://github.com/spatie/better-types now, as it requires `illuminate/collections:^8.58`, but I've also opened a PR to address this issue https://github.com/spatie/better-types/pull/2.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,11 +12,9 @@ jobs:
         strategy:
             matrix:
                 php: [8.0, 8.1]
-                laravel: [8.*, 9.*]
+                laravel: [9.*]
                 dependency-version: [prefer-lowest, prefer-stable]
                 include:
-                    - laravel: 8.*
-                      testbench: ^6.23
                     - laravel: 9.*
                       testbench: 7.*
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,11 +12,13 @@ jobs:
         strategy:
             matrix:
                 php: [8.0, 8.1]
-                laravel: [8.*]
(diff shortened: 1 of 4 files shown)


erikgaal:
Looks like there's an incompatibility between `laravel/framework@8` and `symfony/serializer@6`. Is there a way that we can constrain Laravel 8 installs with Symfony Serializer 5?

brendt:
Yeah, you can add an `exclude` section just like the `include` in the tests CI file.

erikgaal:
Hi @brendt, that will indeed make the CI pass. But what I would like to prevent is someone on L8 without a dependency on `symfony/serializer@v5` installing this package causing Serializer v6 to be installed. Those developers will then run into the same error that cause the CI to fail.

brendt:
Hmm… I don't think we can enforce that on the composer file of this package though…

erikgaal:
Any ideas how we should proceed?

* Option A: Add exclude to CI so build passes. Add documentation to require symfony/serializer v5 in user's projects.
* Option B: Release new version with support only for L9.
* Other options?

freekmurze:
I wouldn't mind a new major that support Laravel 9 only.

brendt:
Let's go with only supporting L9 as a new major release. @erikgaal could you update the PR accortingly?

erikgaal:
PR is ready for review! @brendt @freekmurze

freekmurze:
Looks good to me! Is there anything you'd still like to do before I tag a new major @brendt ?

PR #312, opened 2022-01-19
Remove re-resolving of handler

Robertbaelde:
As per this comment, in V5 the issue that caused the need for re-resolving was fixed.
This PR removes that temporary fix in favor of `Projectionist::fake()` calls. 

@brendt I can imagine this is a breaking change for a small percentage of users, so would require a new major release. I don't think there is a branch for v6 already? 

https://github.com/spatie/laravel-event-sourcing/discussions/181

### DIFF
--- a/src/Projectionist.php
+++ b/src/Projectionist.php
@@ -278,22 +278,6 @@ class Projectionist
 
     private function callEventHandler(EventHandler $eventHandler, StoredEvent $storedEvent): bool
     {
-        /**
-         * We "refresh" an event handler every time it's called, to ensure that its dependencies are properly re-injected.
-         * The underlying problem is with tests: if we're faking an injected dependency when running tests and the handler is already resolved beforehand,
-         * we won't get those faked dependencies.
-         *
-         * A better solution is to store event handler class names instead of an instantiated version of them in the list of handlers, which requires a larger and complex refactor.
-         *
-         * More info here: https://github.com/spatie/laravel-event-sourcing/discussions/181
-         *
-         * Note that we provided `Projectionist::fake` to counter this issue, but it turned out to be very cumbersome to use in complex examples,
-         * which is why we reverted back to the original idea.
-         *
-         * @var \Spatie\EventSourcing\EventHandlers\EventHandler $eventHandler
-         */
-        $eventHandler = app($eventHandler::class);
-
         try {
             $eventHandler->handle($storedEvent);
         } catch (Exception $exception) {


brendt:
I don't think this issue was actually fixed:

> Note that we provided `Projectionist::fake` to counter this issue, but it turned out to be very cumbersome to use in complex examples, which is why we reverted back to the original idea.

Also, keep in mind that testing fakes is only part of the problem we're solving here, you can read about it [over here](https://github.com/spatie/laravel-event-sourcing/discussions/181). Besides testing, we also don't want all handlers to be instantiated when added to the projectionist, that has serieus performance implications. Let me clarify:

imagine an application with 500 domain events, with an average of 2 handlers per event (in reality the average is often much higher); that gives us 1000 handlers that would all be immediately instantiated with every request that comes in, even though a request usually only triggers a handful of handlers. I believe the underlying issue for this PR is your live-replay package you [described here](https://github.com/spatie/laravel-event-sourcing/discussions/181#discussioncomment-1999720)?

In it, you said the following: 

> In order for this to work, the projector needs to know if it's replaying to the live projection or a background projection. For this, a property is set on the instantiated projector.

I don't think we can justify re-introducing a breaking change that would negatively affect both testing and performance. Maybe an easier solution is to keep track of projector state in your package using a dedicated class or data store? Maybe Laravel's cache would even suffice?

Robertbaelde:
> Note that we provided Projectionist::fake to counter this issue, but it turned out to be very cumbersome to use in complex examples, which is why we reverted back to the original idea.

Again, that is a heuristic that your tests are doing way too much. Test components in isolation.

> Also, keep in mind that testing fakes is only part of the problem we're solving here, you can read about it over here. Besides testing, we also don't want all handlers to be instantiated when added to the projectionist, that has serieus performance implications. Let me clarify:

In that case, this PR is actually a performance improvement. What's slower than resolving 1000 handlers at boot of your application? loading them at boot **and** re-resolving them for each event they handle. 

I would agree with the argument when this would re-introduce loading components at application start. 

Up until now, this package seems not too concerned with performance. Especially since on every event we record, we write twice to a database. (In order to set an incremental ID to a meta data field 🤯)

Robertbaelde:
> I don't think we can justify re-introducing a breaking change that would negatively affect both testing and performance.

* This PR will improve performance ;) 
* For testing you could always introduce a "reploadAllHandlers" method. More than happy to add that. 

> Maybe an easier solution is to keep track of projector state in your package using a dedicated class or data store? Maybe Laravel's cache would even suffice?

How would that work? Because when a replay in the background is active, newly recorded events are still recorded and handled. Its not that all projectors must switch at once. Only the projectors that are part of the "replay" should write to the replay db.

brendt:
> Again, that is a heuristic that your tests are doing way too much. Test components in isolation.

FWIW: I don't agree. It should be "test components in isolation AND test them from an integration point of view", in which case this functionality is definitely useful. 

> In that case, this PR is actually a performance improvement. What's slower than resolving 1000 handlers at boot of your application? loading them at boot and re-resolving them for each event they handle.

It's slower to resolve all 1000 handlers at boot than it is to resolve 100 during runtime within the context of a given request. 

> Up until now, this package seems not too concerned with performance.

All the more reasons not to take a step back again. I'm trying my best to improve these issues with every release. I plan on reworking our projectionist implementation from the ground up to further improve the situation.

> How would that work?

You wrote this in [the original issue](https://github.com/spatie/laravel-event-sourcing/discussions/181#discussioncomment-1999720):

> In order for this to work, the projector needs to know if it's replaying to the live projection or a background projection. For this, a property is set on the instantiated projector.
>
> Of course, re-resolving the projector from the container resets its state, breaking the package. I've tried to find a work-around on this, but can't find a good one.

So you need to know what environment you're projecting in. Why do you need to keep that information on the projector itself? Whether you're currently projecting in the live or background environment could be represented by some kind of dependency itself that's injected into your projectors.

```php
class ProjectionEnvironment 
{
    public function isLive(): bool { /* … */ }
}

abstract class AbstractProjector
{
    public function __construct(
        protected ProjectionEnvironment $environment,
    ) {}
}
```

I didn't look at the package in question so I might be missing something; but I reckon there are other ways to determine what kind of environment you're projecting in than to keep a boolean on an event handler instance. From my understanding of what you're trying to do, this shouldn't be a concern of event handlers themselves.

Robertbaelde:
> It's slower to resolve all 1000 handlers at boot than it is to resolve 100 during runtime within the context of a given request.

Yeah, thats true! At the moment, it does both. Which is definitely slower than either of them. I'd understand your argument when the package would not resolve them currently at boot. And this PR would introduce that.

> So you need to know what environment you're projecting in

Let me try to be more clear on what I try to accomplish. Would be great if you could help me find the right way to implement it. 

Imagine we have an application with a large event stream and replaying events takes 10 minutes. Since the replay order matters, we cant accept to record new events during a replay. Also during a replay, your read models potentially are in an incorrect state. 
So when we replay using the default way we have to bring our application down. 

Some applications have strict uptime requirements, and can't be down for 10 mins for a replay. This is where zero-downtime replays come in. 

This is what a zero-downtime replay looks like: 

1. We make a copy of the tables that we replay to (`bank_saldo` table would be `replay_bank_saldo` for example) 
2. We replay to the copy of the table 
3. After this, we keep replaying from the latest replayed event. Until the replay is up to speed with our production events
4. We hot-swap the tables. (rename `bank_saldo` to `bank_saldo_old` and `replay_bank_saldo` to `bank_saldo`). 


The problem I'm facing is that projectors called by the projectionist that is responsible for the replay need to know they are handling a replay while at the same time projectors that are called by the projectionist to record a new event handle it as regular.

I can only think of three ways to accomplish this: 
1. Having a resolved concrete projector in the projectionist
2. add a property to the signature of the handle method
3. In the package completely copy the code of the projectionist

Could you maybe help me find a better way of doing this?

brendt:
I'm happy to think along! Here are a couple of thoughts:

- My approach would be to configure a separate database connection that uses table prefixes (https://laravel.com/docs/8.x/database#read-and-write-connections)
- Replaying one or more projectors should involve a loop that keeps replaying events as long as new events are coming in, **with that separate database connection**
- Projectors themselves should be unaware of which database connection (and thus which prefix) they are using, Projectors that are running in production keep using the default DB connections, while the once you are running via the replay command are using the prefixed connection.
- At some point in time, you'll hot swap your live projections with the newly rebuilt ones. This not only involves table renames, but also a code deploy: you've got your old projectors still running in production, so either you already have versioned events in place and projectors that can handle both of them; or you have a new version of your projectors alongside the old ones. Either way, you need to swap versioned events or projectors at some point. 

To me, that swapping of versioned events or projectors seems like the more difficult part, since the table prefix can be configured without projectors knowing about them, and I'm curious to hear your thoughts on how you'll handle that. I'm also curious to know whether table prefixes actually solve the issue you originally addressed with this PR?

FYI Greg Young wrote a very good book on dealing with event versioning: https://leanpub.com/esversioning, it covers live replay strategies as well.

brendt:
Addressing the performance issue:

> Yeah, thats true! At the moment, it does both. Which is definitely slower than either of them. I'd understand your argument when the package would not resolve them currently at boot. And this PR would introduce that.

It's my plan to get rid of the first initialisation, but it requires more work. That's why getting rid of the on-the-fly initialisation and relying again on the one-time initialisation feels like a step backwards.

Issue #270, opened 2021-10-05
createdAt not available on first run from in projector.


ampersand:
I am running into an issue where the `$event->createdAt()` is returning `null` from the projector handler on the first run.  If the events are replayed the created at value is available.  Is this the intended behavior?

```php
class UserProjector extends Projector
{
    public function onUserUpdated(UserUpdated $event)
    {
        dd($event->createdAt()); // dumps "null" on first run and Carbon date on replay
    }
}
```

Digging through the code I find the following.

https://github.com/spatie/laravel-event-sourcing/blob/f67d9e715ce4dc49887ab9ff8cc5ab8ce11a1736/src/StoredEvents/StoredEvent.php#L114-L118

if I add 
https://github.com/spatie/laravel-event-sourcing/blob/f67d9e715ce4dc49887ab9ff8cc5ab8ce11a1736/src/StoredEvents/StoredEvent.php#L144

before the return the meta data is available on the first run.

PR #324, opened 2022-03-02
Resolves issue with meta data updating on original event.

aidan-casey:
As noted in #270 and #271, there is a bug with the `EloquentStoredEvent` model that causes meta data changes to not be passed on to projectors and reactors the first time the event handled. Subsequent replays will correctly pass the meta data since the "cached" original event is re-created based on meta data in the database. This was a bug originally introduced in #158.

This PR provides a test for the expected functionality and a fix for the bug. The current approach being to check if the `meta_data` attribute is dirty and update the original event if so. While we could add a check for other attributes on the model, I don't think we gain as much value by doing so since those should basically be immutable.

**Side note:** I removed the unused Psalm script from the composer file since it does not appear Psalm is listed as a dependency, and no workflow runs it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,6 @@
         ]
     },
     "scripts": {
-        "psalm": "vendor/bin/psalm",
         "test": "vendor/bin/phpunit",
         "test-coverage": "vendor/bin/phpunit --coverage-html coverage"
     },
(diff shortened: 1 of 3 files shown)


brendt:
Merged: https://github.com/spatie/laravel-event-sourcing/releases/tag/7.0.1

PR #326, opened 2022-03-02
Adds query helpers for event properties

aidan-casey:
This PR adds two methods `wherePropertyIs()` and `wherePropertyIsNot()` to the stored event query builder. These allow you to perform lookups against event properties and utilize the `whereJsonContains` methods. As an example, we have had to use this when querying for all events for a company.

**Example:**

```php
EloquentStoredEvent::query()
    ->whereEvent(CompanyOptedOutOfEmails::class)
    ->wherePropertyIs('companyId', 'ABC-123-XYZ')
    ->get();
```

### DIFF
--- a/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
+++ b/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
@@ -37,7 +37,7 @@ class EloquentStoredEventQueryBuilder extends Builder
         return $this;
     }
 
-    public function whereProperty(string $property, mixed $value): self
+    public function wherePropertyIs(string $property, mixed $value): self
     {
         $this->whereJsonContains(column: "event_properties->{$property}", value: $value);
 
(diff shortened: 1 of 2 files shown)


brendt:
Great addition. Two questions:

- Should we use `whereProperty` or `wherePropertyIs` ? The PR uses the first, but your comment mentions the latter. I prefer the latter one.
- Can you add a test for this?

aidan-casey:
1. Good catch. My intention had been `wherePropertyIs` for consistency, but as you mention that's not reflected in the PR. I'll update that today.

2. Definitely. 👍🏼

aidan-casey:
@brendt - Updated with changes. Let me know what you think.

brendt:
This is good, I'll release it together with https://github.com/spatie/laravel-event-sourcing/pull/327

brendt:
Tagged: https://github.com/spatie/laravel-event-sourcing/releases/tag/7.2.0

PR #331, opened 2022-03-25
Update Projector docs to reflect changes to getting the aggregate uuid

RobHarveyDev:
Following the docs as they are currently causes your projector to not run. Took me a while to figure out this was the reason so felt the docs need to be updated.


freekmurze:
Thank you!

PR #332, opened 2022-03-26
Docs: createdAt() method is camel case

inmanturbo:
This simply fixes a typo. The `createdAt()` method on `ShouldBeStored` is actually `camel case` and not `snake case` like the column name.


freekmurze:
Thanks!

PR #336, opened 2022-04-01
Automatically resolve stored event class alias.

lloricode:


brendt:
I understand what you're trying to do, but could you nevertheless provide a description for future reference?

I'm also not a big fan of the `Helper` class. I wonder if you  considered other alternatives?

lloricode:
Hi @brendt , will check and think of another idea

PR #350, opened 2022-06-15
Fix method annotations in Projection class

daniser:
[According to phpDocumentor docs](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/method.html) correct placement of return type is before method name, like in C. Checked in PhpStorm - it didn't worked with trailing return type.
Also, both `EloquentStoredEventQueryBuilder` and `EloquentStoredEvent` have correct return type placement in their magic method declarations.


freekmurze:
Thanks!

Issue #341, opened 2022-05-07
AggregateRoot Method is called if Argument has no type-hint


fabianpnke:
PHP Version: 8.1.5
Laravel Version: 9.11.0
spatie/laravel-event-sourcing Version: 7.2.0

Hi there,

Today I came across the Situation where Method `two` was getting called when using the AggregateRoot with Method `one` inside the Application. I wondered why this was happening cause I had never used the AggregateRoot with Method `two` anywhere yet. After some experimenting and debugging, I found out that this only happens if the Argument in Method `two` is not type-hinted.

Here are some examples for a better understanding:

Scenario 1 (`two` is called):
```php
<?php

class SomethingAggregateRoot extends AggregateRoot
{
    public function one(string $foo): self
    {
        ...
        $this->recordThat(new SampleEvent());
        ...
        return $this;
    }

    public function two($without_typehint): self
    {
        ...
        $this->recordThat(...);
        ...
        return $this;
    }
}

SomethingAggregateRoot::retrieve($uuid)
    ->one('laravel')
    ->persist();
// Result: two was called
```

Scenario 2 (`two` is not called):
```php
<?php

class SomethingAggregateRoot extends AggregateRoot
{
    public function one(string $foo): self
    {
        ...
        $this->recordThat(new SampleEvent());
        ...
        return $this;
    }

    public function two(boolean $with_typehint): self
    {
        ...
        $this->recordThat(...);
        ...
        return $this;
    }
}

SomethingAggregateRoot::retrieve($uuid)
    ->one('laravel')
    ->persist();
// Result: two was not called
```

Is this an expected behaviour or a bug/defect?

Thanks for checking and clarifying,
Fabian

freekmurze:
To my knowledge, our packages wouldn't call a custom method like `two`.
Could you create a minimal Laravel app in which the problem is reproduced?

fabianpnke:
Please see https://github.com/fabianpnke/laravel-event-sourcing-issue-341-demo

I've managed to reproduce a behaviour in that the problem results in a 500 Server error by removing a type hint from the method.

mariolorente:
Hi,

same here, when there is a public method in an aggregate with non typed params the app totally breaks.

The break occurs exactly inside the apply function of AggregateRoot.php, in the Handlers section.

mariolorente:
mixed doesnt work as a type either.

zackrowe:
[...]

PR #354, opened 2022-07-19
Update AggregateRoot::apply() to utilize acceptsTypes() function

zackrowe:
This PR addresses the bug discovered in https://github.com/spatie/laravel-event-sourcing/issues/341 by implementing the fix I introduced in the `better-types` repository https://github.com/spatie/better-types.

The existing tests should cover that functionality still works as expected. If you would like me to add any additional tests can you point me to the test class they should be applied to?


freekmurze:
Thanks!

PR #362, opened 2022-08-22
Update event-sourcing:replay command to work with just one aggregate uuid

soarecostin:
This MR adds an `--aggregate-uuid=` option to the `event-sourcing:replay` command. When used, the replay command will only replay events for that specific aggregate root uuid.

For context, we use this in order to speed up the replaying process, by running multiple jobs in parallel, each job replaying one specific aggregate uuid.


freekmurze:
Thanks!

PR #363, opened 2022-08-22
Update Reactor docs on how to get the aggregate uuid

soarecostin:
The method stated in the docs for getting the aggregate uuid inside a reactor is not up to date with the latest changes. In fact, adding a second parameter to the reactor will cause it to not fire at all.


freekmurze:
Thank you!

PR #365, opened 2022-09-12
Support weight property to event handlers

sebastiandedeyne:
This is a lightweight (🥁) solution to allow package users to tweak the order projectors and reactors will run.


freekmurze:
Thanks!

PR #369, opened 2022-10-09
Adds abstract ProjectionFactory for creating projections in tests

SolStis86:
Add abstract class ProjectionFactory to allow creation of projection models easily with factory definitions.


freekmurze:
Thanks for this, but I'm not going to merge it. I like to keep this package as small as it can be.

Add this to your own project.

Ahrengot:
I think having a link to this solution in the docs would be helpful. I can add a PR if you'd like.

PR #381, opened 2022-12-22
Use the snapshot with the highest ID

27pchrisl:
If two snapshots are created within the same second, the retrieval method will fetch the first one, not the last one.

Any remaining events are still applied afterwards so the aggregate is consistent, but it means the snapshot effect didn't completely encapsulate the state at the time as expected.

The test confirms that no events needed to be applied after the snapshot is retrieved.


freekmurze:
Thanks!

PR #384, opened 2023-01-03
Fixes bug with retrieving last event

aidan-casey:
This pull request fixes a bug with retrieving the last event when two events exist with the same timestamp. We ran into this issue when two events occurred at the exact same time and the event query failed to pull the proper information in.


freekmurze:
Thanks!

aidan-casey:
No problem!

PR #394, opened 2023-03-01
Proposal: Aliasing events using EventAlias attribute

karkowg:
Hey guys,

I've been using the package in a project and we're basically aliasing all our events in the config file. On top of the `event_class_map` having 150+ lines (and growing) at this point, I have always the concern someone might forget to add an alias there when introducing a new event. It also adds some cognitive load to remember we need to go in the event-sourcing.php file to check what the alias for a certain event is. So, I was thinking why not bring that closer to the source, the event class itself? That way if we open an event class it's already in our faces the fact it has an alias or not. Since we already have a few attributes (version and serializer) we can use in events, I thought it might be a good idea to add a new EventAlias attribute for this purpose.

So your event alias definition would look like this:

```php
#[EventAlias('money_added')]
class MoneyAdded extends ShouldBeStored
{
}
```

Again, a pro with using this approach is we've that information right there with the event definition, so no need to navigate to another file to discover it. Maybe a con is there could be more than one event with the same alias value, although this could also happen in the config approach by declaring the same key in array.

I've done a "quick and dirty" implementation just to get a feeling if the feature would be useful to more people. There's probably a performance concern to be revisited in `StoredEvent::getActualClassForEvent` where we need to find the event class from the alias value, which is not as straightforward as just looking it up in the config map. I've basically got all declared classes and filtered the ones which extend StoredEvent, then filtering the ones that have the attribute, just to get something working.

The implementaion is backwards compatible, so the config map still works, and if a EventAlias is available it takes precedence over the mapped value.

Anyway, thoughts?

### DIFF
--- a/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
+++ b/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
@@ -5,11 +5,10 @@ namespace Spatie\EventSourcing\StoredEvents\Repositories;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\LazyCollection;
-use ReflectionClass;
 use Spatie\EventSourcing\AggregateRoots\Exceptions\InvalidEloquentStoredEventModel;
-use Spatie\EventSourcing\Attributes\EventAlias;
 use Spatie\EventSourcing\Enums\MetaData;
 use Spatie\EventSourcing\EventSerializers\EventSerializer;
+use Spatie\EventSourcing\Facades\Dictionary;
 use Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEvent;
 use Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEventQueryBuilder;
 use Spatie\EventSourcing\StoredEvents\ShouldBeStored;
@@ -131,23 +130,7 @@ class EloquentStoredEventRepository implements StoredEventRepository
 
     private function getEventClass(string $class): string
     {
-        $alias = collect(config('event-sourcing.event_class_map', []))
-            ->flip() // a given event should be aliased only once, so let's flip the collection to make class name the key
-            ->merge(
-                // merge alias from class' EventAlias attribute, if present
-                collect([new ReflectionClass($class)])
-                    ->reject(fn (ReflectionClass $reflection) => empty($reflection->getAttributes(EventAlias::class)))
-                    ->mapWithKeys(function (ReflectionClass $reflection) {
-                        /** @var ReflectionClass<EventAlias> $attribute */
-                        $attribute = $reflection->getAttributes(EventAlias::class)[0];
-
-                        return [$reflection->getName() => ($attribute->newInstance())->alias];
-                    })
-            )
-            ->flip() // flip it back so the alias is the key
-            ->search($class, true);
-
-        return $alias ?: $class;
+        return Dictionary::getAliasFromClass($class) ?: $class;
     }
 
     public function getLatestAggregateVersion(string $aggregateUuid): int
--- a/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
+++ b/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
@@ -5,7 +5,9 @@ namespace Spatie\EventSourcing\StoredEvents\Repositories;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\LazyCollection;
+use ReflectionClass;
 use Spatie\EventSourcing\AggregateRoots\Exceptions\InvalidEloquentStoredEventModel;
+use Spatie\EventSourcing\Attributes\EventAlias;
 use Spatie\EventSourcing\Enums\MetaData;
 use Spatie\EventSourcing\EventSerializers\EventSerializer;
 use Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEvent;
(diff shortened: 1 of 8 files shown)


mcadet:
@freekmurze I honestly think it's a good idea.

Thank you for your awesome packages at Spatie!

src/StoredEvents/Repositories/EloquentStoredEventRepository.php
```diff
@@ -129,13 +131,23 @@ public function update(StoredEvent $storedEvent): StoredEvent
...
-        if (! empty($map) && in_array($class, $map)) {
-            return array_search($class, $map, true);
-        }
-
-        return $class;
+        $alias = collect(config('event-sourcing.event_class_map', []))
+            ->flip() // a given event should be aliased only once, so let's flip the collection to make class name the key
+            ->merge(
+                // merge alias from class' EventAlias attribute, if present
+                collect([new ReflectionClass($class)])
```
freekmurze, line 138: Let's make this more readable, by retrieving and merging in the event alias in a separate function. By naming that new function right, you could lose this comment.

freekmurze: Same remarks for the other changes, try to break it up in smaller functions (that will also make it easier for users to override stuff)

freekmurze (changes requested):
I like the functionality, let's polish it a bit more.

src/StoredEvents/Repositories/EloquentStoredEventRepository.php
karkowg: Thanks Freek, I will revisit this as soon as possible.

src/StoredEvents/StoredEvent.php
```diff
@@ -149,17 +150,46 @@ protected function instantiateEvent(?ShouldBeStored $originalEvent): void
 
     protected static function getActualClassForEvent(string $class): string
     {
-        return Arr::get(config('event-sourcing.event_class_map', []), $class, $class);
+        /** @var LazyCollection<string,string> $aliasedClasses */
+        $aliasedClasses = LazyCollection::make(function () {
+            foreach (get_declared_classes() as $declaredClass) {
```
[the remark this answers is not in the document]
karkowg: This logic has now been moved to https://github.com/spatie/laravel-event-sourcing/pull/394/commits/f0104188efc11fce7f55a8c0e4b9aedcdcb9bf38#diff-0dff908a5ae8127a4d53e526951b5c90d4e3202eca8071ec2f047a0abb383127R17

src/StoredEvents/Repositories/EloquentStoredEventRepository.php
karkowg: @freekmurze I've extracted a Dictionary facade since some of the logic was duplicated between StoredEvent and EloquentStoredEventRepository. Let me know if it makes sense.

daniser:
@karkowg 
> The implementaion is backwards compatible, so the config map still works, and if a EventAlias is available it takes precedence over the mapped value.

I think it should be the other way around - aliases in config file should have precedence over "hardcoded" alias values. For example, I have events in my package with their "default" alias values, but when I install package I have to remap them to avoid possible conflicts with another events.

karkowg:
> @karkowg 
> > The implementaion is backwards compatible, so the config map still works, and if a EventAlias is available it takes precedence over the mapped value.
> 
> I think it should be the other way around - aliases in config file should have precedence over "hardcoded" alias values. For example, I have events in my package with their "default" alias values, but when I install package I have to remap them to avoid possible conflicts with another events.

Not sure I understand your concern. Could you give an example?

Btw if you never define an alias using the EventAlias attribute, it would continue using the one in the config file. In my opinion you would use either the config file or the attribute for aliasing a specific event, not both. Or none of them to have the "default" alias.

Also don't think using the attribute strategy is any more "hardcoded" than the config file since both are optional ways of mapping a class name to a string.

Hope that helps.

daniser:
@karkowg The difference is that the config file is publishable, and events are not. That's why I've called `EventAlias`'d aliases "hardcoded". I myself working on package using event sourced approach provided by this package. Feature you are working on is very demanded, but there may be (very little) probability that "hardcoded" event aliases already occupied by another events in the application in which I want to install my package. And, from the perspective of package user, I cannot access those aliases and change them inside that package.
So it'd be good addition to your feature if the end user would have the ability to change those aliases in their published `event-sourcing.php` config file.

karkowg:
> @karkowg The difference is that the config file is publishable, and events are not. That's why I've called `EventAlias`'d aliases "hardcoded". I myself working on package using event sourced approach provided by this package. Feature you are working on is very demanded, but there may be (very little) probability that "hardcoded" event aliases already occupied by another events in the application in which I want to install my package. And, from the perspective of package user, I cannot access those aliases and change them inside that package. So it'd be good addition to your feature if the end user would have the ability to change those aliases in their published `event-sourcing.php` config file.

@daniser I see. Well, now with the [Dictionary](https://github.com/karkowg/laravel-event-sourcing/blob/f0104188efc11fce7f55a8c0e4b9aedcdcb9bf38/src/Dictionary.php) class (pending review), we could make it so you can [provide your own implementation](https://github.com/karkowg/laravel-event-sourcing/blob/f0104188efc11fce7f55a8c0e4b9aedcdcb9bf38/src/EventSourcingServiceProvider.php#L68). This way you would be able to customize what takes precedence or even have different aliasing logic.

That said, I don't mind changing the precedence to the config file but will let the maintainers decide what is best for the package.

sebastiandedeyne:
Sorry for barging is so late, but I'd like to take a step back and consider another option: a (static) `name` method on `ShouldBeStored`.

```php
class MoneyAdded extends ShouldBeStored
{
    public static function name(): string
    {
        return 'money_added';
    }
}
```

I think this has a few benefits over attributes. The only drawback I can think of is it doesn't look as good as a single line class attribute :) (which we could actually support easier with this set up)

1) You can programmatically customize the event name

If you want more compact event names based on the class name, create your own base `ShouldBeStored` class with a custom `name` implementation.

```php
abstract class ShouldBeStored extends BaseShouldBeStored
{
    // …
    
    public static function name(): string
    {
        return Str::snake(class_basename($this));
    }
}
```

2) You can enforce implementation

If you want enforce all events to have an explicit name, create your own base `ShouldBeStored` class and throw an exception in the method.

```php
abstract class ShouldBeStored extends BaseShouldBeStored
{
    // …
    
    public static function name(): string
    {
        throw new Exception('Name not implemented');
    }
}
```

3) No discussion about config/event class precedence

The name would _always_ be driven by the event class. The base implementation would read config.

(Example implementation)

```php
abstract class ShouldBeStored
{
    // …
    
    public static function name(): string
    {
        return collect(config('event-sourcing.event_class_map'))
            ->flip()[static::class] ?? static::class;
    }
}
```

PR #398, opened 2023-03-14
Make projections trait based instead of class based

filippotoso:
Hi,

if you use a code generator to build your base models, having projections based on a class makes it difficult to reuse the code.

This PR moves the projections' feature in a trait and an interface. Existing code will keep working. 

For the users who want to extends their models class into projections, they just need to add an interface and a trait:

```php
class UserProjection extends Model implements Writable
{
    use Projectionable;
    // ...
}
```


freekmurze:
I'm going to leave things as they are for now to keep things simple. 

I'll revisit this when creating a new major version of the package.

PR #401, opened 2023-03-16
Make commands Lazy to allow extending

EriBloo:
Hi,

this small pull makes all commands Lazy loaded to support overriding. This makes it similiar to Laravel commands. Commands can also be Lazy through `$defaultName` property, but since this syntax is deprecated I didn't add it.


EriBloo:
Ok. This pull is a bit redundant.

I thought that `AsCommand` is essential for Laravel to know which command to use, but it actually isn't. The problem I had while using laravel-event-sourcing was that I added this Attribute on my classes extending base generators. In a situation like this Laravel would pick base event-sourcing commands.

This commit is not crucial then, but it can ensure no matter what strategy users choose (with or without AsCommand Attribute) proper command will be resolved. I'm not deleting this pull and would like for you to decide if that is something helpful.

freekmurze:
Thanks, I'm passing on this one for now to keep things simple.

Issue #396, opened 2023-03-07
Docs - versions cross linked 


Shkeats:
Thanks for the package!

I was confused when navigating the docs site that I kept ending up with v5 as the selected version despite starting on v7.

<img width="292" alt="image" src="https://user-images.githubusercontent.com/2048674/223413333-5e999fc3-f0e9-49eb-9160-71299b168893.png">

It seems that some of the hardcoded links in the docs take you back to old versions (including v4) of the docs and this can cause a bit of confusion as the nav updates on the site and then any subsquent browsing will keep you on that old version, potentially leading you to read the wrong stuff!

Would be a fairly straightforward find/replace PR but maybe there's a more sustainable way to fix this?

<img width="1017" alt="image" src="https://user-images.githubusercontent.com/2048674/223413997-71166b36-93cd-4f67-9caf-2963b8b0806c.png">

<img width="1006" alt="image" src="https://user-images.githubusercontent.com/2048674/223413942-abcb77cb-55be-4300-b5e6-f4b1220baa44.png">

freekmurze:
Could you send a PR on the branches where these version numbers are wrong?

PR #402, opened 2023-03-16
Fix documentation version cross-linking to other versions for v7. #396

Shkeats:
Fix documentation version cross-linking to other versions for v7. #396

All of the updated links/files seem to exist in the current branch too so no edits needed.

If we're happy with this one, I can pick up tidying up the older version branches too.


freekmurze:
Thanks!

PR #403, opened 2023-03-19
Fix README.md version cross-linking to other versions for v7. #396

Shkeats:
For #396 I missed the outdated links in the README.md as I had constrained my search to the docs folder.

See: https://github.com/spatie/laravel-event-sourcing/pull/402


freekmurze:
Thanks!

PR #405, opened 2023-03-31
Fix preserving event order when dispatched from aggregate root via event handler

daniser:
When firing event from reactor, dispatched event should go after event that triggered the reactor. Example:
`MoneyAdded -> DoubleBalanceReactor -> MoneyMultiplied`
currently dispatches events to the bus `MoneyMultiplied` first. This PR fixes that behaviour.

### DIFF
--- a/src/StoredEvents/StoredEvent.php
+++ b/src/StoredEvents/StoredEvent.php
@@ -64,14 +64,12 @@ class StoredEvent implements Arrayable
 
     public function handleForAggregateRoot(): void
     {
-        $this->handle();
-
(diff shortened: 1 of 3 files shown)


freekmurze:
@rubenvanassche I think this is fine. Could you double check that this won't result in issues for our client projects?

rubenvanassche:
@sebastiandedeyne, doesn't look like a problem to me, even a solution to some problems, what do you think?

sebastiandedeyne (approved):

sebastiandedeyne:
Looks good, thanks!

PR #406, opened 2023-04-23
Fix snapshots in aggregate partials

sonja-turo:
This pull request is an attempted, updated fix for issue #307 where the states of partials are not restored when a snapshot is retrieved.

This update works by defining a single key (default `__esPartials`) in the event sourcing config file that is used as a key in the state of a snapshot. Each partial is then reflected for public properties, and a key that is the classname of the partial is added to the `__esPartials` in the generated. The value of that key is the state of the partial.

This will generate a snapshot state of:
```php
$state = [
    'foo' => 'bar',
    '__esPartials' => [
        'App/Partials/FooPartial' => [
            'fooProperty1' => 1,
            'fooProperty2' => true
    ],
];
````

Reconstitution of the snapshot goes through each key of __esPartials and matches to a resolved Partial Class, and sets the state of that Partials.

### DIFF
--- a/config/event-sourcing.php
+++ b/config/event-sourcing.php
@@ -75,6 +75,13 @@ return [
      */
     'snapshot_model' => Spatie\EventSourcing\Snapshots\EloquentSnapshot::class,
 
+    /**
+     * The key within an AggregateRoot under which snapshots of partials are stored.
+     * This is stored within the Aggregate state when snapshotted, and restored when
+     * a snapshot is retrieved.
+     */
+    'snapshot_partials_key' => '__esPartials',
+
     /*
      * This class is responsible for handling stored events. To add extra behaviour you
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by sebastiandedeyne** [comment]
Looks good! One thing I'd consider changing is removing the partials key altogether. Slashes can't be used for PHP properties anyway, so there won't be a clash in property names. Might as well keep it flat then.

```php
$state = [
    'foo' => 'bar',
    'App\Partials\FooPartial' => [
        'fooProperty1' => 1,
        'fooProperty2' => true
    ],
];
```

sonja-turo:
The unneeded _esPartials key has been removed.

sonja-turo:
@sebastiandedeyne Just a little bump.

sebastiandedeyne:
Thanks!

PR #407, opened 2023-04-30
`ReplayCommand` `--stored-event-model` option is no longer implemented

Shkeats:
ReplayCommand `--stored-event-model` option is no longer implemented and the value isn't accessed in the command. As far as I can tell this is no longer needed as the custom model can be specified in config and the Eloquent repo respects that config value.

The test covering it is already skipped, so I assume this was just missed cleanup.

This PR, removes the option from the command, the reference in the docs and the skipped test.


avosalmon:
I have implemented it coincidentally https://github.com/spatie/laravel-event-sourcing/pull/410 haha.
I believe this option is quite useful!

sebastiandedeyne:
I ended up merging @avosalmon's PR :) Thanks!

PR #409, opened 2023-05-19
Document the use of the Projection class

27pchrisl:
Discovered this helpful class in the Event Sourcing in Laravel book which isn't documented in the `docs`, so I have added it here.


freekmurze:
Thanks!

PR #410, opened 2023-06-01
Replay events with custom stored event model

avosalmon:
ReplayCommand `--stored-event-model` option is not implemented even though it's [documented](https://spatie.be/docs/laravel-event-sourcing/v7/advanced-usage/replaying-events).

> If you are [using your own event storage model](https://spatie.be/docs/laravel-event-sourcing/v7/advanced-usage/using-your-own-event-storage-model/) then you will need to use the --stored-event-model option when executing event-sourcing:replay to specify the model storing the events you want to replay.

This option would be especially useful when the application stores events in different tables depending on the domain. In this scenario, specifying the custom model in the config doesn't work since the custom model will be dynamically determined based on the domain.
Related discussion: https://github.com/spatie/laravel-event-sourcing/discussions/347

### DIFF
--- a/src/Console/ReplayCommand.php
+++ b/src/Console/ReplayCommand.php
@@ -31,9 +31,7 @@ class ReplayCommand extends Command
             return;
         }
 
-        if ($this->hasOption('stored-event-model')) {
-            $model = $this->option('stored-event-model');
-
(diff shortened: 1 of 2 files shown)


sebastiandedeyne:
Thanks!

PR #412, opened 2023-06-05
dont overwrite created_at if already set

Ahrengot:
This introduces the ability to manually set the "created_at" for events when dispatching them through aggregates.

This is useful, especially during the migration of an existing application to event sourcing, where it is necessary to preserve the "created at" metadata.

I encountered this need myself while writing a console command that resembles the code snippet below, and realised that the `recordThat()` method would overwrite the "created at" value I had specified.

```PHP
SomeModel::cursor()->each(function(SomeModel $original) {
    $event = new SomethingHappened();
    $event->setCreatedAt($original->created_at->toImmutable());

    SomeAggregate::retrieve(...)
                 ->recordThat($event)
                 ->persist();
});
```

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -105,7 +105,7 @@ abstract class AggregateRoot
     {
         $domainEvent
             ->setAggregateRootUuid($this->uuid)
-            ->setCreatedAt(CarbonImmutable::now());
+            ->setCreatedAt($domainEvent->createdAt() ?? CarbonImmutable::now());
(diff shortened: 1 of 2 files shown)


sebastiandedeyne:
I tend to keep any data that affects business logic part of the event payload, `created_at` isn't available there. If the date something happened is important, I use a specific property like `paidAt` or `shippedAt`.

Second, if you override `created_at`, you lose sight on when the event was added to the system, which can be valuable information when debugging. Treat `created_at` as technical metadata, not for application logic use.

PR #416, opened 2023-06-14
Correcting Commands Documentation

ChangingTerry:
Updated to show that commands not longer implement `Spatie\Shop\Support\EventSourcing\Commands\Command` interface


sebastiandedeyne:
Thanks!

PR #424, opened 2023-07-20
Support recording events concurrently

sebastiandedeyne:
This is a proof of concept PR that allows events to be stored concurrently,  avoiding the dreaded `CouldNotPersistAggregate` exception in specific circumstances.

This can be useful to do work in parallel, for example on a queue. Currently, if two jobs process the same aggregate root (AR) simultaneously, one of the jobs will fail because of the `CouldNotPersistAggregate` exception. Or worse, when an AR is being processed in a queued job while a web request is processing the same AR, the user might receive a 500 server error.

A common case we've had is when an AR manages multiple "child" aggregates. For example, an `OrderAggregateRoot` might contain state for multiple order items. We currently cannot process the order items simultaneously because the shared AR creates a bottleneck. The goal of this PR is to allow work to be done concurrently under specific circumstances, while maintaining the current behaviour to avoid ARs to end up in an invalid state.

There are no changes to the current behavior in aggregate roots. If you want to opt into concurrently persisting events, you must replace `recordThat` calls with `recordConcurrently` and `persist` calls with `persistConcurrently`. Ideally this becomes the default behavior in a unified API in the future, but that would be a breaking change that requires a major version tag.

```diff
  class AccountAggregateRoot extends AggregateRoot
  {
      private int $balance = 0;

      public function addMoney(int $amount): void
      {
-         $this->recordThat(new MoneyAdded($amount));
+         $this->recordConcurrently(new MoneyAdded($amount));
      }

      protected function applyMoneyAdded(MoneyAdded $event): void
      {
          $this->balance += $event->amount;
      }
}
```

(In a future major version, I'd rather have an `allowConcurrent` argument on `recordThat`, but that would be a breaking change.)

If multiple processes call `AccountAggregateRoot::addMoney` simultaneously, they'll both be persisted instead of throwing an `CouldNotPersistAggregate` exception.

You may also provide a callback to `recordConcurrently` for additional validation.

```php
class AccountAggregateRoot extends AggregateRoot
{
    // …

    public function removeMoney(int $amount): void
    {
        $this->recordConcurrently(
            new MoneyRemoved($amount),
            function (self $accountAggregateRoot) {
                if ($accountAggregateRoot->balance < $amount) {
                    throw new Exception('Insufficient balance');
                }
            }
        );
    }

    protected function applyMoneyRemoved(MoneyRemoved $event): void
    {
        $this->balance -= $event->amount;
    }
}
```

In this example, `MoneyRemoved` can only be recorded if the balance is high enough. Here's what happens when `persistConcurrently` is called:

- First, we'll run the concurrency check (so you only need to validate your action once)
- We'll attempt to persist the event
- If the `persist` method throws a `CouldNotPersistAggregate` exception, we'll re-retrieve the aggregate root in its latest state and re-record the new events. While we re-record, we'll pass the aggregate root through all concurrency checks again to ensure it still is in a valid state

This way we can allow aggregate roots to be persisted in parallel while guarding against invalid states when needed.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -34,14 +34,12 @@ abstract class AggregateRoot
 
     protected int $aggregateVersionAfterReconstitution = 0;
 
+    protected ?self $aggregateInstanceRetrievedAfterReconstitution = null;
+
     /** @var \Illuminate\Support\Collection|\Spatie\EventSourcing\AggregateRoots\AggregatePartial[] */
     protected Collection $entities;
 
     private bool $handleEvents = true;
-
-    private int $concurrentTries = 0;
-
-    private int $maximumTries = -1;
 
     /** @var callable[] */
     private array $concurrencyChecks = [];
@@ -58,8 +56,11 @@ abstract class AggregateRoot
         $aggregateRoot = app(static::class);
 
         $aggregateRoot->uuid = $uuid;
+        $aggregateRoot->aggregateInstanceRetrievedAfterReconstitution = clone $aggregateRoot;
 
-        return $aggregateRoot->reconstituteFromEvents();
+        $aggregateRoot->reconstituteFromEvents();
+
+        return $aggregateRoot;
     }
 
     public function loadUuid(string $uuid): static
@@ -137,6 +138,19 @@ abstract class AggregateRoot
 
     public function persist(): static
     {
+        $storedEvents = $this->persistWithoutApplyingToEventHandlers();
+
+        if ($this->handleEvents) {
+            $storedEvents->each(fn (StoredEvent $storedEvent) => $storedEvent->handleForAggregateRoot());
+        }
+
+        $this->aggregateVersionAfterReconstitution = $this->aggregateVersion;
+
+        return $this;
+    }
+
+    public function persistConcurrently(int $tries = -1): static
+    {
         $recordedEvents = $this->recordedEvents;
         $concurrencyChecks = $this->concurrencyChecks;
 
@@ -153,16 +167,11 @@ abstract class AggregateRoot
         try {
             $storedEvents = $this->persistWithoutApplyingToEventHandlers();
         } catch (CouldNotPersistAggregate $exception) {
-            // @todo Add tests for maximum tries
-            if ($this->maximumTries !== -1 && $this->concurrentTries >= $this->maximumTries) {
+            if ($tries === 1) {
                 throw $exception;
             }
 
-            // @todo Instead of creating a new instance, this should reset the aggregate root state back to it's
-            //       initial state after __construct, then re-retrieve events from the database and apply them
-            //       on the resetted aggregate root.
             $newInstance = static::retrieve($this->uuid());
-            $newInstance->concurrentTries = $this->concurrentTries + 1;
 
             foreach ($recordedEvents as $i => $recordedEvent) {
                 $concurrencyCheck = $concurrencyChecks[$i];
@@ -176,8 +185,7 @@ abstract class AggregateRoot
                 $newInstance->recordConcurrently($recordedEvent, $concurrencyCheck);
             }
 
-            // @todo Don't return a new instance but the same one, see above todo
-            return $newInstance->persist();
+            return $newInstance->persistConcurrently($tries === -1 ? -1 : ($tries - 1));
         }
 
         if ($this->handleEvents) {
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -138,6 +138,16 @@ abstract class AggregateRoot
 
     public function persist(): static
     {
+        foreach ($this->recordedEvents as $i => $recordedEvent) {
+            $concurrencyCheck = $this->concurrencyChecks[$i];
+
+            if (is_null($concurrencyCheck)) {
+                continue;
+            }
+
+            $concurrencyCheck($this);
+        }
+
         $storedEvents = $this->persistWithoutApplyingToEventHandlers();
 
         if ($this->handleEvents) {
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -158,6 +158,9 @@ abstract class AggregateRoot
                 throw $exception;
             }
 
+            // @todo Instead of creating a new instance, this should reset the aggregate root state back to it's
+            //       initial state after __construct, then re-retrieve events from the database and apply them
+            //       on the resetted aggregate root.
             $newInstance = static::retrieve($this->uuid());
             $newInstance->concurrentTries = $this->concurrentTries + 1;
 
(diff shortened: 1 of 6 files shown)


src/AggregateRoots/AggregateRoot.php
```diff
@@ -116,6 +125,25 @@ public function recordThat(ShouldBeStored $domainEvent): static
         return $this;
     }
 
+    public function recordConcurrently(ShouldBeStored $domainEvent, bool|callable $allowConcurrent = true): static
```
brendt, line 128: What's the use case for being able to pass a boolean to `$allowConcurrent`.

It's kind of confusing to me that you'd call `recordConcurrently`, and then pass `false` to force it not to be concurrent.

src/AggregateRoots/AggregateRoot.php
```diff
@@ -129,6 +157,42 @@ public function persist(): static
...
 
+    public function persistConcurrently(int $tries = -1): static
+    {
+        $recordedEvents = $this->recordedEvents;
+        $concurrencyChecks = $this->concurrencyChecks;
+
+        try {
+            $storedEvents = $this->persistWithoutApplyingToEventHandlers();
+        } catch (CouldNotPersistAggregate $exception) {
+            if ($tries === 1) {
```
brendt, line 168: Shouldn't this be `0` instead of `1`?

src/AggregateRoots/AggregateRoot.php
```diff
@@ -129,6 +157,42 @@ public function persist(): static
...
     }
 
+    public function persistConcurrently(int $tries = -1): static
+    {
+        $recordedEvents = $this->recordedEvents;
+        $concurrencyChecks = $this->concurrencyChecks;
+
+        try {
+            $storedEvents = $this->persistWithoutApplyingToEventHandlers();
+        } catch (CouldNotPersistAggregate $exception) {
```
brendt, line 167: One concern: are there any ways `CouldNotPersistAggregate` is thrown in cases where you can't recover from retrying? We should be sure you can't end up in an infinite loop.

src/AggregateRoots/AggregateRoot.php
```diff
@@ -129,6 +157,42 @@ public function persist(): static
         return $this;
     }
 
+    public function persistConcurrently(int $tries = -1): static
```
brendt, line 160: Would it be possible to use `persist()` as a shorthand if you don't want to specify tries? That way there'll be less mental overhead for the common use case.

src/AggregateRoots/AggregateRoot.php
```diff
@@ -116,6 +125,25 @@ public function recordThat(ShouldBeStored $domainEvent): static
         return $this;
     }
 
+    public function recordConcurrently(ShouldBeStored $domainEvent, bool|callable $allowConcurrent = true): static
+    {
```
brendt, line 129: One idea I had for a potential API to make it a bit more easy, is to not have a separate `recordConcurrently` method, but instead make the `record` method able to accept a special object `ConcurrentEvent`. You could add a helper method to convert a `ShouldBeStored` to such a special `ConcurrentEvent` object, it could simply be a wrapper as well, if you want to keep things simpler.

```php
$ar->recordThat(concurrent(new MoneyAdded(…)))
```

Just an idea :)

src/AggregateRoots/AggregateRoot.php
sebastiandedeyne: Ideally the signature would be `true|callable` to allow concurrency without a check, but that doesn't exist in PHP 🙃 So instead I supported booleans in general, which means we have to handle false too (even though it'd never be used)

I agree this is confusing, I updated the signature to `?callable`, when `$allowConcurrent` is `null` we won't do any extra checks.

src/AggregateRoots/AggregateRoot.php
sebastiandedeyne: This is what I was building initially. However, when the aggregate root is re-retrieved after a concurrency exception, it has a new reference because it's a new object. Unfortunately there's no way to keep the original AR object without heavy use of reflection and black PHP magic.

That would be a breaking change for the `persists` method. This could cause issues:

```php
$aggregateRoot = AggregateRoot::retrieve(…);
// …

$aggregateRoot->persist();
$aggregateRoot->doSomethingElse();
```

To fix:

```php
$aggregateRoot = AggregateRoot::retrieve(…);
// …

$aggregateRoot = $aggregateRoot->persist();
$aggregateRoot->doSomethingElse();
```

I think this is too risky of a change to introduce in a minor version. In the next major version, I'd like to address this and provide a single API for concurrently recording events (both for `record`/`recordConcurrently` and `persist`/`persistConcurrently`

src/AggregateRoots/AggregateRoot.php
sebastiandedeyne: Think I'm going to hold off and keep the API separate for now. In the next major version I want a single API for all as described above. Storing a different type of object internally also isn't easy without changing the contents of `$recordedEvents`, which could be considered a breaking change.

src/AggregateRoots/AggregateRoot.php
sebastiandedeyne: Hmm I can't think of any other places… Do you think we should not allow `-1` for tries and have a hard limit of `100`?

src/AggregateRoots/AggregateRoot.php
brendt: FYI `true` is a standalone type in 8.2: https://stitcher.io/blog/new-in-php-82#null,--true,-and-false-as-standalone-types-rfc

sebastiandedeyne: Good to know! Still supporting 8.1 for now :)

src/AggregateRoots/AggregateRoot.php
brendt: It might be a good failsafe? 100 seems good enough, I cannot imagine anyone wanting to retry a 100 times, and if they want to, they can still overwrite the value. But better to be safe than sorry, I think?

src/AggregateRoots/AggregateRoot.php
brendt: Gotcha, it makes sense then to keep them separate :)

bjarn:
Hi @sebastiandedeyne,

I stumbled on this PR some time ago and was curious what you found out and why you did not finish. If you have any info, let me know! I would be very keen to work on this myself, but looking for some info to get a headstart here 🙌

sebastiandedeyne:
This branch was pretty much complete, but I decided not to merge it because we didn't end up needing it and it added a lot of complexity under a different API. I want to have this in the package, but shoehorning it into the current major version felt like a band-aid solution. I'd rather revisit this in a next major package version, and have it part of the main API.

sebastiandedeyne:
Yeah we essentially took a step back and a totally different approach to the problem that didn't require concurrency (sorry, been too long and too specific to remember the exact details)

inmanturbo:
@sebastiandedeyne thank you :). That's all I was wondering, not specific details but if you just avoided concurrency altogether. Good to know👍

PR #430, opened 2023-08-10
Ignore abstract event handlers

27pchrisl:
This PR filters out abstract projectors and reactors, so that application code can use abstract subclasses.


sebastiandedeyne:
Thanks!

PR #433, opened 2023-08-15
Make the aggregatePartial recordThat method fluent

maartenpaauw:
Just like the `recordThat` on the `aggregateRoot` abstract class.


freekmurze:
Thank you!

PR #438, opened 2023-09-01
Added an extra rejection to prevent events being passed to invalid handlers

### LINKED ISSUE #437: Version 7.3.7 tries to apply events to variadic and untyped functions in the aggregate root
Hi,

The change made in https://github.com/spatie/laravel-event-sourcing/pull/434 appears to be causing the apply() method to call any methods with variadic or mixed/non-declared types on the aggregate root, as it believes they can handle the event.

```php
        Handlers::new($this)
            ->public()
            ->protected()
            ->reject(fn (Method $method) => in_array($method->getName(), ['handleCommand', 'recordThat', 'apply', 'tap']))
            ->accepts($event) // Captures anything in the aggregate root that can receive this type
            ->all()
            ->each(function (Method $method) use ($event) {
                return $this->{$method->getName()}($event);
            });
```

I found this because I was using mocks for aggregate roots, and mockery adds a lot of these types of methods. I quickly wrote this which implements a rejection for method that have a variadic parameter, no type, or the type isn't either an interface or is_a the event.

If this is along the right lines I'm happy to tidy it up into a PR?

```php
        Handlers::new($this)
            ->public()
            ->protected()
            ->reject(fn (Method $method) => in_array($method->getName(), ['handleCommand', 'recordThat', 'apply', 'tap']))
            ->reject(fn (Method $method) => collect(
                (new \ReflectionClass($this))->getMethod($method->getName())->getParameters())
                ->filter(
                    fn (\ReflectionParameter $parameter) => !$parameter->isVariadic() && $parameter->getType() && (interface_exists($parameter->getType()->getName()) || is_a($event, $parameter->getType()->getName(), true))
                )
                ->isEmpty())
            ->all()
            ->each(function (Method $method) use ($event) {
                return $this->{$method->getName()}($event);
            });
```

27pchrisl:
Fixes #437. Essentially if you'll accept null you're too accepting (un-typed params, variadics etc.) I also added a test to check handling interfaces still works.

### DIFF
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -256,6 +256,7 @@ abstract class AggregateRoot
             ->public()
             ->protected()
             ->reject(fn (Method $method) => in_array($method->getName(), ['handleCommand', 'recordThat', 'apply', 'tap']))
+            ->reject(fn (Method $method) => $method->accepts(0))
             ->accepts($event)
             ->all()
             ->each(function (Method $method) use ($event) {
--- a/src/AggregateRoots/AggregateRoot.php
+++ b/src/AggregateRoots/AggregateRoot.php
@@ -73,6 +73,7 @@ abstract class AggregateRoot
             ->public()
             ->protected()
             ->reject(fn (Method $method) => in_array($method->getName(), ['handleCommand', 'recordThat', 'apply', 'tap']))
+            ->reject(fn (Method $method) => $method->accepts(null))
             ->accepts($command)
             ->first()
         ) {
@@ -86,6 +87,7 @@ abstract class AggregateRoot
                 ->public()
                 ->protected()
                 ->reject(fn (Method $method) => in_array($method->getName(), ['recordThat', 'apply', 'tap']))
+                ->reject(fn (Method $method) => $method->accepts(null))
                 ->accepts($command)
                 ->first();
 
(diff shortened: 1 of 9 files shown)


src/AggregateRoots/AggregateRoot.php
```diff
@@ -256,6 +256,7 @@ protected function apply(ShouldBeStored $event): void
             ->public()
             ->protected()
             ->reject(fn (Method $method) => in_array($method->getName(), ['handleCommand', 'recordThat', 'apply', 'tap']))
+            ->reject(fn (Method $method) => $method->accepts(null))
```
sebastiandedeyne, line 259: Very elegant fix!

sebastiandedeyne (changes requested):

sebastiandedeyne (changes requested):
Can we do this in `HandlesEvents` too to keep the logic between discovering handlers for aggregate roots & projectors/reactors consistent?

Or could that be a breaking change for some?

27pchrisl:
I can't think of a reason someone would have deliberately created a real event handler that could accept null as its first argument, or even done that accidentally....

sebastiandedeyne:
The only case I can think of is a "catch-all" projector or reactor without a typehint, but in that case it can be fixed by hinting `ShouldBeStored`.

Merging and tagging, thank you!

PR #446, opened 2023-11-13
Document prepare events using abstract class instead of interface

Sparclex:
Currently the documentation tells you to implement an non-exsting interface. This change updates the documentation according to the code.


freekmurze:
Thanks!

PR #447, opened 2023-11-24
Change the column type of snapshots.aggregate_version

eschalks:
Just started using this package for the first time for a new project. 
Noticed the two migrations that are included have a different column type for what appears to be the same data. (As the snapshot's aggregate_version always seems to match the corresponding event's aggregate_version).
I decided to use the bigger data type of the two as this seemed like the safer option.


sebastiandedeyne:
Thanks!

PR #458, opened 2024-03-08
L11 compatibility

inmanturbo:
[symfony/serializer](https://github.com/symfony/serializer) made the `Symfony\Component\Serializer\Normalizer\ObjectNormalizer` class final and now requires to implement some of the core methods. 

We still only needed to change `getSupportedTypes()` so rather than copy in whole function bodies I just grabbed an instance of their `ObjectNormalizer` and used it for the rest.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,52 +1,59 @@
 name: run-tests
 
 on:
-    push:
-    pull_request:
-    schedule:
-        - cron: '0 0 * * *'
+  push:
+  pull_request:
+  schedule:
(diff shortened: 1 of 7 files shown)


inmanturbo:
@freekmurze if you prefer another pattern for implementing the stubs, or want to name things differently, please feel free to edit or request changes.

inmanturbo:
Note on the forced pushed:

I had squashed everything down but I though you might want to see the commits that came from running the shift, so I brought back the full history

freekmurze:
Thank you very much!

PR #459, opened 2024-03-15
AllowDynamicProperties to suppress PHP warnings on 8.2

### LINKED ISSUE #435: Deprecation notice: $firedFromAggregateRoot is deprecated
Using PHP 8.2, I am getting this deprecation notice, if I set `'dispatch_events_from_aggregate_roots' => true,`:

>DEPRECATED  Creation of dynamic property ...  $firedFromAggregateRoot is deprecated in vendor/spatie/laravel-event-sourcing/src/StoredEvents/StoredEvent.php on line 68.

Should be possible to allow dynamic props or define them?

sebastiandedeyne:
Fixes #435

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "spatie/laravel-package-tools": "^1.9",
         "spatie/laravel-schemaless-attributes": "^2.0",
         "symfony/finder": "^6.0|^7.0",
+        "symfony/polyfill-php82": "^1.29",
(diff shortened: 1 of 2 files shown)


daniser:
@sebastiandedeyne attribute added to the wrong class. Should be added to `ShouldBeStored` instead.

freekmurze:
This has been fixed!

PR #461, opened 2024-03-29
Add config to enforce every event has an alias.

thettler:
Like with `Relation::enforceMorphMap` i added a config option to always throw an exception if an event without alias is persisted. 

Thank you for the great package if you have questions or change requests I am happy to update the PR.


freekmurze:
Thanks for this high quality PR! 👍

PR #463, opened 2024-05-13
Pass specified aggregate uuid to resetState while replaying

meijdenmedia:
When using the --aggregate-uuid while replaying, I would like the resetState method to be informed about the parameter so I can reset only a part of the projector.


freekmurze:
Thanks!

Issue #414, opened 2023-06-06
Attribute EventSerializer is not used for serialize


philipp-kern:
Hello!

Using the PHP attribute \Spatie\EventSourcing\Attributes\EventSerializer it is possible to set a specific serializer for an event.
This works well for deserializing the event, but not on serialize.

Because \Spatie\EventSourcing\EventSerializers\EventSerializer defines serialize in its interface I would consider this a bug.

Please also use the specified serializer on serialize of the events.

PR #464, opened 2024-05-26
Use custom serializer while persisting events

sebastiandittrich:
I stumbled upon issue #414 while setting up custom event serializers.

This PR fixes this issue by also using the custom event serializer while persisting events.

If you have any questions, please reach out to me!

## Note About Versioning
I don't know if the previous behavior was a bug or intentional, but "fixing" it changes how events are serialized into the database. Therefore, it probably makes sense to release this in a major version update.

### DIFF
--- a/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
+++ b/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
@@ -5,10 +5,14 @@ namespace Spatie\EventSourcing\StoredEvents\Repositories;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\LazyCollection;
+use ReflectionClass;
+use ReflectionException;
 use Spatie\EventSourcing\AggregateRoots\Exceptions\InvalidEloquentStoredEventModel;
+use Spatie\EventSourcing\Attributes\EventSerializer as EventSerializerAttribute;
 use Spatie\EventSourcing\Enums\MetaData;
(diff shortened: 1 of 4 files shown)


src/StoredEvents/Repositories/EloquentStoredEventRepository.php
```diff
@@ -83,8 +87,20 @@ public function persist(ShouldBeStored $event, string $uuid = null): StoredEvent
...
 
+        try {
+            $reflectionClass = new ReflectionClass(get_class($event));
+        } catch (ReflectionException) {
+            throw new InvalidStoredEvent();
+        }
+
+        if ($serializerAttribute = $reflectionClass->getAttributes(EventSerializerAttribute::class)[0] ?? null) {
+            $serializerClass = ($serializerAttribute->newInstance())->serializerClass;
+        } else {
```
freekmurze, line 98: We really dislike `else` statements. Could you refactor this way by putting this line before the if statement?

```
$serializerClass = EventSerializer::class
```

freekmurze (changes requested):
Could you add a test that proves that change works?

src/StoredEvents/Repositories/EloquentStoredEventRepository.php
sebastiandittrich: Sure!

sebastiandittrich:
@freekmurze I added a test, let me know if this works for you 😄

freekmurze:
Thanks!

PR #468, opened 2024-06-12
fix: internal links use /docs prefix

rburgt:
Internal links in hosted documentation link to 404 as `/docs` prefix is missing in linked url. This change adds the `/docs` prefix to all internal links.


sebastiandedeyne:
Thanks!

PR #471, opened 2024-06-20
add failing test for serializing `created_at` event property

Ahrengot:
Hi there!

I encountered this bug when I was trying to store a collection of Eloquent models as events (Long story...)

If you define a stored event property with the name `created_at`, it will be set to null when the event is serialiazed.

I've been digging through the code, trying to figure out why this happens and found the following:

The ObjectNormalizer uses the PropertyAccessor from Symfony, and setting the following line (**In the Symfony package**) to false will fix the behavior:
https://github.com/symfony/property-access/blob/7.1/PropertyAccessor.php#L472

The ObjectNormalizer is triggering the PropertyAccessor on this line:
https://github.com/spatie/laravel-event-sourcing/blob/main/src/Support/ObjectNormalizer.php#L55

Do you know if there is a way to configure this from outside? I'm open to contributing a fix myself, but I'm a little out of my comfort zone, and I don't want to reinvent any wheels unnecessarily 😄

### DIFF
--- a/tests/EventSerializers/EventSerializerTest.php
+++ b/tests/EventSerializers/EventSerializerTest.php
@@ -8,6 +8,7 @@ use DateTimeImmutable;
 
 use function PHPUnit\Framework\assertEquals;
 use function PHPUnit\Framework\assertInstanceOf;
+use function PHPUnit\Framework\assertStringContainsString;
 
 use Spatie\EventSourcing\EventSerializers\EventSerializer;
 use Spatie\EventSourcing\Tests\TestClasses\Events\EventWithArray;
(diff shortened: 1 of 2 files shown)


Ahrengot:
I think I understand what's going on now. 

The magic getter in PropertyAccessor is capturing this method from `ShouldBeStored`:
https://github.com/spatie/laravel-event-sourcing/blob/main/src/StoredEvents/ShouldBeStored.php#L27-L30

Is there any way around that?

sebastiandedeyne:
Unfortunately I don't think so. I wrestled with similar issues with the Symfony package in the past but didn't get far.

I'm not sure this is something we can/should fix now. This seems like a niche issue that would disappear in a next major version by not having Symfony's serializer as default, and improving how we deal with created at dates because there are a few inconsistencies that need to be ironed out.

PR #473, opened 2024-06-26
Update installation-setup.md to match actual contents of config/event-sourcing.php

inmanturbo:
I just copied from the published config. The one at https://spatie.be was showing `use` statements at the top of the file and they were all formatted into a single line without breaks.

The actual config just uses fully qualified class names.


sebastiandedeyne:
Thanks!

Issue #84, opened 2020-04-15
The replay command should have a --force flag


sebastiandedeyne:
To be consistent with other Laravel commands, running `php artisan event-sourcing:replay --force` should run without a confirmation prompt.

pascualstromsnes:
Hello @sebastiandedeyne 

I have made a PR (#86) that includes an implementation of a force flag and a corresponding test.

freekmurze:
Thanks @pascualstromsnes, we'll continue this in #86

PR #474, opened 2024-07-07
add force option to replay command

inmanturbo:
#84
#86

### DIFF
--- a/src/Console/ReplayCommand.php
+++ b/src/Console/ReplayCommand.php
@@ -13,7 +13,8 @@ class ReplayCommand extends Command
     protected $signature = 'event-sourcing:replay {projector?*}
                             {--from=0 : Replay events starting from this event number}
                             {--stored-event-model= : Replay events from this store}
-                            {--aggregate-uuid= : Replay events for this aggregate only}';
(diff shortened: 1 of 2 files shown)


src/Console/ReplayCommand.php
```diff
@@ -45,7 +46,7 @@ public function handle(Projectionist $projectionist): void
     public function selectProjectors(array $projectorClassNames): ?Collection
     {
         if (count($projectorClassNames) === 0) {
-            if (! $this->confirm('Are you sure you want to replay events to all projectors?', true)) {
+            if ($this->isRunningInteractively() && ! $this->confirm('Are you sure you want to replay events to all projectors?', true)) {
```
sebastiandedeyne, line 49:
```suggestion
            if (! $this->option('force') && ! $this->confirm('Are you sure you want to replay events to all projectors?', true)) {
```

I don't think this is worth extracting to a method.

sebastiandedeyne (changes requested):
Thanks! 🙏 Minor question, we can merge after.

inmanturbo:
I've committed your changes. Thanks for taking a look!

sebastiandedeyne:
Thanks, tagged and released!

PR #476, opened 2024-08-15
Use class map name when querying event class

maartenpaauw:
When using an `event_class_map` and querying `->whereEvent(...)`, all given events should be mapped to their corresponding value before querying the database.

### DIFF
--- a/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
+++ b/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
@@ -3,6 +3,8 @@
 namespace Spatie\EventSourcing\StoredEvents\Models;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Arr;
+use Spatie\EventSourcing\StoredEvents\StoredEvent;
 
 /**
  * @method \Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEventCollection get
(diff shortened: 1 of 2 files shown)


src/StoredEvents/StoredEvent.php
```diff
@@ -150,7 +150,7 @@ protected static function getActualClassForEvent(string $class): string
         return Arr::get(config('event-sourcing.event_class_map', []), $class, $class);
     }
 
-    protected static function getEventClass(string $class): string
+    public static function getEventClass(string $class): string
```
maartenpaauw, line 153: Unfortunately I had to make this method public, because I don't want to duplicate it's logic.

maartenpaauw:
Some clarification: I opened this pull request because the documentation suggests passing the class name to the `->whereEvent(...)` method. You can find more information here: https://spatie.be/docs/laravel-event-sourcing/v7/advanced-usage/event-queries

src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
```diff
@@ -32,7 +34,10 @@ public function whereAggregateRoot(string $uuid): self
 
     public function whereEvent(string ...$eventClasses): self
     {
-        $this->whereIn('event_class', $eventClasses);
+        $this->whereIn('event_class', array_map(
+            static fn (string $eventClass): string => StoredEvent::getEventClass($eventClass),
```
sebastiandedeyne, line 38:
```suggestion
           fn (string $eventClass): string => StoredEvent::getEventClass($eventClass),
```

I assume for performance or an IDE default? But it's negligible in this case, would rather have it without.

sebastiandedeyne:
Thanks!

PR #484, opened 2024-09-06
Only Dispatch If Event Has Async Handler

damiantw:
This PR modifies the condition under which the handle stored event job is dispatched to the queue. 

With the pre-existing behavior, a job is dispatched each time an event persists if the projectionist has any async handlers registered, even if the specific event is irrelevant to all async handlers.

My application has certain events that contain a lot of data. These events do not have async handlers. Serializing these events to the queue (for what is ultimately a noop job) has resulted in errors due to SQS payload limits.

I also have some high-volume events that do not have async handlers, leading to unnecessary excessive queue throughput.

This change ensures a job is only dispatched if the fired event has async handlers.


sebastiandedeyne:
Thanks!

PR #485, opened 2024-09-20
Replace 'projector' with 'reactor' where necessary.

PHPGuus:
Seems like the creating-and-configuring-reactors,md file was created by copying the same file from the using-projectors directory, and some instances of "projector" were left in tact without being replaced by "reactor".

This PR fixes that documentation issue.


riasvdv:
Thanks!

PR #490, opened 2024-10-11
update command name to match class name

PH7-Jack:
### Description
This PR aims to fix a problem in the `EventSourcingServiceProvider:54`, where the clear command is typed differently from the command signature.

Screenshot 2024-10-10 at 22 43 39

### Fixed Error when running `optimize:clear`
Screenshot 2024-10-10 at 22 44 15

### DIFF
--- a/src/EventSourcingServiceProvider.php
+++ b/src/EventSourcingServiceProvider.php
@@ -51,7 +51,7 @@ class EventSourcingServiceProvider extends PackageServiceProvider
         if (method_exists($this, 'optimizes')) {
             $this->optimizes(
                 optimize: 'event-sourcing:cache-event-handlers',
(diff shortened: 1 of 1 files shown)


sebastiandedeyne:
Hmm I think this fix should be in EventSourcingServiceProvider:54, not the command. Otherwise it's a breaking change.

sebastiandedeyne (changes requested):
Hmm I this fix should be the other way around: EventSourcingServiceProvider:54 should be updated, not the command signature. Otherwise it's a breaking change.

PH7-Jack:
@sebastiandedeyne Sure, it makes sense. I've made the change.

sebastiandedeyne:
Thanks!

PR #493, opened 2024-11-28
PHP 8.4 explicit null types

BertvanHoekelen:
Hi, this PR removes the deprecation notices regarding the implicitly nullable parameters.


riasvdv:
Thanks!

PR #495, opened 2024-12-20
Fix Remaining Nullable Parameter Definitions

okaufmann:
Fixed the remaining missing ?string definitions for the metadata parameter in EventSerializer, JsonEventSerializer, and the documentation example for consistency.


freekmurze:
Thanks!

PR #496, opened 2025-01-21
feat: add eloquent stored event generic types for collection and query builder

maartenpaauw:
This is useful when using PHPStan in your project in combination with the `EventQuery` classes.

### DIFF
--- a/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
+++ b/src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
@@ -6,7 +6,11 @@ use Illuminate\Database\Eloquent\Builder;
 use Spatie\EventSourcing\StoredEvents\StoredEvent;
 
 /**
- * @method \Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEventCollection get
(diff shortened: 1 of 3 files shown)


src/StoredEvents/Models/EloquentStoredEventQueryBuilder.php
```diff
@@ -6,7 +6,11 @@
 use Spatie\EventSourcing\StoredEvents\StoredEvent;
 
 /**
- * @method \Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEventCollection get
+ * @template TEloquentStoredEvent of EloquentStoredEvent
+ *
+ * @method \Spatie\EventSourcing\StoredEvents\Models\EloquentStoredEventCollection<EloquentStoredEvent> get()
```
maartenpaauw, line 11: I tried to pass `TEloquentStoredEvent` as a type to the `EloquentStoredEventCollection`, but unfortunately, it is not possible. The best alternative solution is to pass `EloquentStoredEvent` as the type instead.

freekmurze:
Thanks!

maartenpaauw:
Thanks for the quick merge and release!

PR #499, opened 2025-03-13
Fix the CarbonImmutable's toJson method to always convert to UTF time

youyingxiang:
The persist code of `Spatie\EventSourcing\StoredEvents\Repositories\EloquentStoredEventRepository`.

```php
public function persist(ShouldBeStored $event, string $uuid = null): StoredEvent
    {
       .....
        $eloquentStoredEvent->setRawAttributes([
            'event_properties' => app($serializerClass)->serialize(clone $event),
            'aggregate_uuid' => $uuid,
            'aggregate_version' => $event->aggregateRootVersion(),
            'event_version' => $event->eventVersion(),
            'event_class' => $this->getEventClass(get_class($event)),
            'meta_data' => json_encode($event->metaData() + [
                MetaData::CREATED_AT => $createdAt->toDateTimeString(),
            ]),
            'created_at' => $createdAt,
        ]);
        .......
    }
```

If I set the timezone in `config/app.php` to a non-UTC timezone, `json_encode` will always set the `created-at` in `meta_data` to the UTC timezone, resulting in inconsistent timestamps.

### DIFF
--- a/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
+++ b/src/StoredEvents/Repositories/EloquentStoredEventRepository.php
@@ -98,13 +98,18 @@ class EloquentStoredEventRepository implements StoredEventRepository
             $serializerClass = $serializerAttribute->newInstance()->serializerClass;
         }
 
+        $metaData = $event->metaData();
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #500, opened 2025-03-29
feat:  Add `ignoringDirectories` to filter `Providers`, `route` directory

youyingxiang:
@freekmurze  The `addToProjectionist` code of `Spatie\EventSourcing\Support\DiscoverEventHandlers`.

```php
public function addToProjectionist(Projectionist $projectionist)
{
    if (empty($this->directories)) {
        return;
    }

    $files = (new Finder())->files()->in($this->directories);
    
    return collect($files)
        ->reject(fn (SplFileInfo $file) => in_array($file->getPathname(), $this->ignoredFiles))
        ->map(fn (SplFileInfo $file) => $this->fullQualifiedClassNameFromFile($file))
        ->filter(fn (string $eventHandlerClass) => is_subclass_of($eventHandlerClass, EventHandler::class))
        ->filter(fn (string $eventHandlerClass) => (new ReflectionClass($eventHandlerClass))->isInstantiable())
        ->pipe(function (Collection $eventHandlers) use ($projectionist) {
            $projectionist->addEventHandlers($eventHandlers->toArray());
        });
}

```

### Case 1

The `addToProjectionist` method scans the entire `app` directory. The composer package `laravel/telescope` is not installed in production environment, but executing `is_subclass_of($eventHandlerClass, EventHandler::class)` results in an error `Class "Laravel\Telescope\TelescopeApplicationServiceProvider" not found`, because `App\Providers\TelescopeServiceProvider` extends `Laravel\Telescope\TelescopeApplicationServiceProvider`.

image

image

image

### Case 2

When scanning all route files in `app/Api/routes/`, `is_subclass_of` triggers autoloading. The autoload mechanism causes routes to be registered. This is equivalent to executing `require app/Api/routes/bookings.php`, which causes `Route::resource('bookings', BookingsController::class)->only(['store', 'destroy'])` to be executed

image

### DIFF
--- a/config/event-sourcing.php
+++ b/config/event-sourcing.php
@@ -11,6 +11,18 @@ return [
     ],
 
     /*
+     * These directories will be ignored when scanning for projectors and reactors.
+     */
+    'auto_discover_projectors_and_reactors_should_ignore_directories' => [
+        app()->path('Providers'),
+        app()->path('Console'),
+        app()->path('Events'),
+        app()->path('Listeners'),
+        app()->path('Jobs'),
+        app()->path('Notifications'),
+    ],
+
+    /*
      * This directory will be used as the base path when scanning
      * for projectors and reactors.
      */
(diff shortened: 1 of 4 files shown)


freekmurze:
Thanks for your work on this.

To keep the package simple, I'm going to leave it as it is now.  If the default config is causing problems, I suggest changing `auto_discover_projectors_and_reactors` in your project.

PR #501, opened 2025-04-14
Case insensitive CouldNotPersistAggregate::invalidVersion detection

nick-potts:
With Postgres, the error message thrown for a unique constraint violation doesn't have a capital D

```
SQLSTATE[  23505]: Unique violation:   7 ERROR:  duplicate key value violates unique constraint 
```


nick-potts:
Would probably make more sense to compare to the UniqueConstraintViolationException

PR #503, opened 2025-04-27
[Docs] Add Missing Version Property in Event Serializers

devhammed:
### [Docs] Add Missing Version Property in Event Serializers

**Context:**  
The documentation for using a custom event serializer was outdated and inconsistent with the updated `EventSerializer` interface.

**Changes Introduced:**  
- Added PHP syntax highlighting to code blocks for better readability.
- Updated the `EventSerializer` interface method signature:
  - `deserialize` now accepts an additional `int $version` parameter.
  - Made `$metadata` optional and nullable.
- Adjusted the example `UpgradeSerializer` class:
  - Updated its `deserialize` method signature to match the new interface.
  - Added a conditional check to apply currency defaulting logic only for the `MoneyAdded` event class.

**Impact:**  
This update ensures developers extending or overriding the event serializer have accurate references, No breaking changes to codebase functionality.

### DIFF
--- a/docs/advanced-usage/using-your-own-event-serializer.md
+++ b/docs/advanced-usage/using-your-own-event-serializer.md
@@ -9,7 +9,7 @@ You can specify your own serializer by creating a class that implements `Spatie\
 
 This is the content of the `EventSerializer` interface:
 
-```
+```php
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #506, opened 2025-06-12
feat: add dynamic weight

Bloemendaal:
Hi! We have the following situation in our code:

```php
class OfferOccupationProjector extends Projector
{
    public int $weight = 4;

    public function onContractCartItemAdded(ContractCartItemAddedEvent $event): void
    {
        (new OfferOccupation)
            ->writeable()
            ->insert([
                'id'                    => Str::uuid()->toString(),
                'contract_cart_item_id' => $event->cartItemId,
                'offer_id'              => $event->offerId,
                'start_date'            => $event->startDate,
                'end_date'              => $event->endDate,
            ]);
    }
}
```

This projector is listening to the `ContractCartItemAddedEvent`, because once that happens, an offer occupation should be made. The weight of this projector is set to 4, because otherwise this projector would run before the `CartProjector`, causing this code to throw an exception because the foreign key constraint on `contract_cart_item_id` fails.

I found that the previous people that worked on this project removed the `OfferOccupation` in the `CartProjector`, which is weird, why not do this in the `OfferOccupationProjector`. Well, it turns out that because of the weight, the removal function below would run after the removal of the cart, meaning the database fails again because of the restrict on the foreign key. This won't work:

```php
class OfferOccupationProjector extends Projector
{
    public int $weight = 4;

    public function onContractCartItemRemoved(ContractCartItemRemovedEvent $event): void
    {
        OfferOccupation::query()
            ->where('contract_cart_item_id', '=', $event->cartItemId)
            ->delete();
    }
}
```

So I know there's probably a better way to implement this with events, but the issue is that I cannot really change this anymore, so I figured: why not make the weight property a function with the event in it. Now, everyone with this same situation can base the order of the projectors (and reactors) on the `StoredEvent` (and `null` when it is resetting the state).

The change to the `getWeight` is completely optional, the `weight` property still works.

In my case, this would be the implementation:
```php
class OfferOccupationProjector extends Projector
{
    public function getWeight(?StoredEvent $event): int
    {
         return match ($event?->event_class) {
              ContractCartItemAddedEvent::class => 4,
              ContractCartItemRemovedEvent::class => -4,
              default => 0,
         };
    }
}
```

### DIFF
--- a/docs/using-projectors/making-sure-events-get-handled-in-the-right-order.md
+++ b/docs/using-projectors/making-sure-events-get-handled-in-the-right-order.md
@@ -36,6 +36,31 @@ class MyProjector extends Projector
 }
 ```
 
+Alternatively, you can determine the weight dynamically based on the event being processed. This allows you to prioritize certain events over others. The `$event` parameter will be `null` when the `resetState()` method is called (during projector reset operations).
+
+```php
+namespace App\Projectors;
+
+use App\Events\MoneyAddedEvent;
+use App\Events\MoneySubtractedEvent;
+use Spatie\EventSourcing\EventHandlers\Projectors\Projector;
+use Spatie\EventSourcing\StoredEvents\StoredEvent;
+
+class MyProjector extends Projector
+{
+    public function getWeight(?StoredEvent $event): int
+    {
+        return match ($event?->event_class) {
+            MoneyAddedEvent::class => 2,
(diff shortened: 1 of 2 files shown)


jordanvanbeijnhem (approved):

freekmurze:
We'd like to merge this in, put could you add tests and docs first?

freekmurze:
Thanks!

PR #509, opened 2025-06-20
Event replay optimizations

suhailhijry:
This PR adds the ability replay only events related to the selected projectors, and uses ```chunk``` to optimize memory usage and avoid memory corruption bugs in PHP when replaying millions of events.


suhailhijry:
Of course, I updated the code now, it should run all the tests correctly.

PR #516, opened 2026-01-20
Introduce configurable aggregate event ordering.

### LINKED ISSUE #515: [MySQL] Out of sort memory
When using MySQL (8.0.35 in our case), loading events for a single aggregate can fail with:

> SQLSTATE[HY001]: Memory allocation error: 1038 Out of sort memory, consider increasing server sort buffer size

This can happen even for aggregates with very few events if one event row contains a large JSON payload.  
The root cause is that events are ordered by `id` instead of `aggregate_version`.

From an event-sourcing perspective, `aggregate_version` is the canonical ordering and already has a unique composite index with `aggregate_uuid`.

## Environment

- **Package**: spatie/laravel-event-sourcing (7.10.3 & 7.12.1 & 7.12.3)
- **Laravel**: 11.44.2 & 11.46.1 & 11.47.0
- **PHP**: PHP 8.2 & PHP 8.3
- **Database**: MySQL 8.0.35
- **Storage engine**: InnoDB
- **sort_buffer_size**: 262144 (256 KB)

One event in the failing aggregate has a JSON payload of ~530 KB in event_properties.

## Failing Query

```
SELECT *
FROM stored_events
WHERE aggregate_uuid = ?
ORDER BY id ASC;
```

## Observed behavior

MySQL chooses the index on (aggregate_uuid, aggregate_version) for the lookup.

* Because the query orders by `id`, MySQL performs a filesort.
* With `SELECT *`, MySQL’s filesort uses addon fields, so the large JSON payload is included in the sort buffer.
* The sort buffer (256 KB) is exceeded by a single large row, resulting in: "Out of sort memory (1038)"

`EXPLAIN FORMAT=TREE` shows:

```
-> Sort: stored_events.id
    -> Index lookup on stored_events using stored_events_aggregate_uuid_aggregate_version_unique
```

Forcing an index on `(aggregate_uuid, id)` makes the query succeed, confirming that the issue is the sort step.

## Expected behavior

Loading events for an aggregate should be stable regardless of event payload size and should not depend on MySQL sort buffer limits.

## Proposed solution

Change aggregate event retrieval to order by `aggregate_version` instead of `id`:

```php
->where('aggregate_uuid', $uuid)
->orderBy('aggregate_version')
```

Reasons:

* `aggregate_version` is the semantically correct ordering for an aggregate’s event stream.
* It matches the existing unique index `(aggregate_uuid, aggregate_version)`.
* It avoids filesorts entirely and eliminates this class of MySQL failures.
* It removes the need for index hints or database-level tuning.

-----------

Would you accept a change to order aggregate events by `aggregate_version` instead of `id`? Would that change be considered a breaking change for a new major release? It's a minor code change with more or less massive impact, it shouldn't change anything about results etc - but changing SQL queries is definitely a candidate for breaking things. But at the same time it's also a bug fix - as the current sort by `id` is semantically incorrect and also results in errors for big aggregates or single big events.

Gummibeer:
fixes #515

### DIFF
--- a/tests/EloquentStoredEventRepositoryTest.php
+++ b/tests/EloquentStoredEventRepositoryTest.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\EventSourcing\Tests;
 
+use PHPUnit\Framework\Assert;
 use function PHPUnit\Framework\assertEquals;
 use function PHPUnit\Framework\assertSame;
 
@@ -58,8 +59,8 @@ it('uses id ordering by default when retrieving aggregate events', function () {
     $repository = new EloquentStoredEventRepository();
     $events = $repository->retrieveAll('uuid-order-test');
 
-    $eventArray = $events->toArray();
-    assertCount(3, $eventArray);
+    $eventArray = $events->all();
+    Assert::assertCount(3, $eventArray);
 
     // Events should be ordered by id (which is the default)
     assertEquals(100, $eventArray[0]->event->amount);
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #519, opened 2026-02-09
Fix metaData leaking into event_properties since Symfony Serializer v8.0.4

### LINKED ISSUE #518: metaData property leaks into event_properties since Symfony Serializer v8.0.4
After upgrading the Symfony Serializer component from v8.0.3 to v8.0.4/v8.0.5, the metaData property from ShouldBeStored is being serialized into the event_properties column of the stored_events table.      
                                                                                                                                                                                                                 
My assumption is that this happens because ShouldBeStored has a public metaData() getter method for the protected $metaData property. The Symfony ObjectNormalizer now recognizes this getter and includes metaData in the serialized output. In Symfony v8.0.3 this was not the case.                                                                                           
                                                                                                                                                                                                                 
The metaData should only be stored in the dedicated meta_data column, not duplicated in event_properties.                                                                                                      
                                                                                                                                                                                                                 
Steps to reproduce:                                                                                                                                                                                            
                                                                                                                                                                                                                 
  1. Install spatie/laravel-event-sourcing v7.12.x with symfony/serializer v8.0.4 or higher                                                                                                                      
  2. Record any event via an AggregateRoot                                                                                                                                                                       
  3. Inspect the event_properties column in the stored_events table                                                                                                                                              
  4. The serialized JSON now contains a metaData key                                                                                                                                                             
                                                                                                                                                                                                                 
Expected behavior:                                                                                                                                                                                             
                                                                                                                                                                                                                 
  metaData should not appear in event_properties, as it was the case with Symfony Serializer v8.0.3.                                                                                                             
                                                                                                                                                                                                                 Workaround:                                                                                                                                                                                                    
                                                                                                                                                                                                                 
Create a custom normalizer that filters out metaData from the extracted attributes and register it in the event_normalizers config

freekmurze:
## Summary
- Fixes `metaData` (and other `ShouldBeStored` internal properties) leaking into the `event_properties` column when using Symfony Serializer v8.0.4+
- Uses Symfony's `#[Ignore]` attribute on all `ShouldBeStored` methods to exclude them from serialization
- Configures `ClassMetadataFactory` with `AttributeLoader` in the custom `ObjectNormalizer` so the `#[Ignore]` attributes are read

Closes #518

## Test plan
- [x] Added test asserting `metaData` is not present in serialized event properties
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/StoredEvents/ShouldBeStored.php
+++ b/src/StoredEvents/ShouldBeStored.php
@@ -7,12 +7,14 @@ use Carbon\CarbonImmutable;
 use ReflectionClass;
 use Spatie\EventSourcing\Attributes\EventVersion;
 use Spatie\EventSourcing\Enums\MetaData;
+use Symfony\Component\Serializer\Attribute\Ignore;
 
 #[AllowDynamicProperties]
 abstract class ShouldBeStored
 {
     protected array $metaData = [];
 
+    #[Ignore]
     public function eventVersion(): int
     {
         $versionAttribute = (new ReflectionClass($this))->getAttributes(EventVersion::class)[0] ?? null;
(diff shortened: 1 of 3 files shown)


freekmurze:
@georgecoca Thanks for reporting this! I've created a fix in #524 that takes a different approach: instead of using `#[Ignore]` attributes (which leak into child classes), the `ObjectNormalizer` now filters out `ShouldBeStored`'s internal attributes directly, while preserving child class properties with the same names.

Would you be able to verify that #524 fixes the issue for your use case?

Issue #181, opened 2020-12-11
Change projector and reactor instantiation


brendt:
Whenever a projector or reactor is added to the projectionist, it's always immediately resolved from the container:

```php
class Projectionist
{
    public function addProjector($projector): Projectionist
    {
        if (is_string($projector)) {
            $projector = app($projector);
        }
```

This approach has a few down sides:

- All event handlers are immediately instantiated, even when not used
- It makes it impossible to swap out reactors and projectors with fakes during testing.

Consider the following test case:

```php
    /** @test */
    public function test_pay_payment()
    {
        FakeOrderPaymentPaidReactor::fake();
        
        Carbon::setTestNow('2020-01-01');

        $payment = PaymentAggregateFactory::new()->create();

        $payment = (new PayPayment)($payment);

        $this->assertTrue($payment->state->equals(PaidPaymentState::class));
        $this->assertTrue(Carbon::create('2020-01-01')->eq($payment->paid_at));
    }
```

This integration test will test for quite a lot of things. One side effect of a payment being paid though is that an invoice is generated with browsershot in the `OrderPaymentPaidReactor`. I of course want a testcase checking whether this invoice generation happens correctly, but in most other test cases this adds unnecessary overhead; like in the previous example. 

`FakeOrderPaymentPaidReactor::fake();` unfortunately doesn't work, since the reactor was already resolved from the container during the framework setup. That's unfortunate and I'd like to change it.

PR #523, opened 2026-02-13
Lazy-resolve event handlers to eliminate boot-time overhead

georgecoca:
When registering projectors and reactors via class name strings, defer container resolution until the handlers are actually needed (event dispatch, replay, or explicit access). This avoids resolving all handlers and their dependency trees on every request, since the vast majority of requests never produce a stored event.

The boot-time app() calls were already redundant — callEventHandler() re-resolves every handler via app() on each invocation anyway (see #181).

Here's a benchmark  **(14 projectors + 24 reactors, 1000 iterations)**:

| Scenario                              | avg     | p50     | p99      |
|---------------------------------------|---------|---------|----------|
| Eager (original — `app()` on boot)    | 62.5 µs | 59.9 µs | 133.6 µs |
| Lazy (new — boot only)                | 4.4 µs  | 4.4 µs  | 5.0 µs   |
| Lazy (boot + resolve on first use)    | 63.7 µs | 62.5 µs | 89.3 µs  |

**Boot is 14× faster (93% reduction).**  
Total cost is deferred, not eliminated, paid on first event dispatch.  
Most HTTP requests never fire a stored event, so the cost is never paid.

In our app we had this improvement:
**Before vs After (averages across 5 runs)**

| Metric                         | Before   | After  | Improvement           |
|--------------------------------|----------|--------|-----------------------|
| Total boot to route            | 153.69ms | 7.63ms | 95% faster (20×)      |
| EventSourcingServiceProvider   | 143.95ms | 0.37ms | 99.7% faster (389×)   |

The EventSourcingServiceProvider provider was consuming **93.7% of total boot time**.  
With lazy loading it drops to **under half a millisecond**.

### DIFF
--- a/src/Projectionist.php
+++ b/src/Projectionist.php
@@ -28,12 +28,37 @@ class Projectionist
 
     protected bool $isReplaying = false;
 
+    /** @var array<string, string> */
+    protected array $pendingProjectors = [];
+
+    /** @var array<string, string> */
+    protected array $pendingReactors = [];
+
     public function __construct(array $config)
     {
         $this->projectors = new EventHandlerCollection();
         $this->reactors = new EventHandlerCollection();
 
         $this->catchExceptions = $config['catch_exceptions'];
+    }
+
+    private function resolve(): void
+    {
+        if (empty($this->pendingProjectors) && empty($this->pendingReactors)) {
+            return;
+        }
+
+        foreach ($this->pendingProjectors as $class) {
+            $this->projectors->addEventHandler(app($class));
+        }
+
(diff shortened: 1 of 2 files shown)


freekmurze:
@georgecoca Great contribution, thanks! The benchmarks speak for themselves — 14x faster boot is a big win. Merged and released as [7.13.0](https://github.com/spatie/laravel-event-sourcing/releases/tag/7.13.0).

I simplified the `addReactor` method to match `addProjector`'s pattern (added `string|EventHandler` type hint and removed redundant `is_string` checks) in a follow-up commit.

